PR #7, opened 2021-11-05
[MO] - KafkaVersionService, Generic containers replaced by builder

see-quick:

This PR adding few changes in `test-container` repo:

1. the First problem was that we add two more instance properties to containers (kafka_version, strimzi_test_container_version), which results in a lot of factory methods... I decided to re-factor the builder pattern and make it more flexible. 
2. the Second problem is that we do not want to have any information about Kafka and Strimzi test container versions. The solution is the class (`KafkaVersionService`), which takes care of this. In case more interested see the `KafkaVersionService` class.

It works on the following JSON schema:
```
{
  "version": 1,
  "kafkaVersions": {
    "3.0.0": "test-container:latest-kafka-3.0.0",
    "2.40.50": "test-2",
    "2.8.1": "test-container:latest-kafka-2.8.1"
     ... (more versions...)
  }
}
```

This PR will be blocked until we decide, where do we want to store JSON file with versions


src/main/java/io/strimzi/utils/LogicalKafkaVersionEntity.java
```diff
@@ -0,0 +1,205 @@
...
+        /**
+         * Compare two decimal version strings, e.g. 1.10.1 &gt; 1.9.2
+         * @param version1 The first version.
+         * @param version2 The second version.
+         * @return 0 if version1 == version2;
+         * -1 if version1 &lt; version2;
+         * 1 if version1 &gt; version2.
+         */
+        public static int compareVersions(String version1, String version2) {
+            String[] components = version1.split("\\.");
```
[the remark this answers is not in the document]
see-quick: help

see-quick:
@scholzj )

.checkstyle/checkstyle.xml
```diff
@@ -87,7 +87,10 @@
         <!-- code quality -->
         <module name="MethodLength"/>
         <module name="ParameterNumber"/>
-        <module name="ClassDataAbstractionCoupling"/>
+        <module name="ClassDataAbstractionCoupling">
+            <!-- default is 7 -->
+            <property name="max" value="8"/>
+        </module>
```
scholzj, line 93: Would annotation be better then overriding it here? It will make sure you don't change it everywhere but only where really needed.

pom.xml
```diff
@@ -96,6 +96,7 @@
         <docker-java.version>3.2.7</docker-java.version>
         <kafka.version>3.0.0</kafka.version>
         <slf4j.version>1.6.1</slf4j.version>
+        <gson.version>2.8.5</gson.version>
```
scholzj, line 99: Why GSON and what is it used for? I think all of the Strimzi projects use Jackson.

src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
```diff
@@ -58,7 +58,8 @@ public StrimziKafkaCluster(final int brokersNum, final int internalTopicReplicat
         this.brokersNum = brokersNum;
         this.network = Network.newNetwork();
 
-        this.zookeeper = new StrimziZookeeperContainer()
+        this.zookeeper = new StrimziZookeeperContainer.StrimziZookeeperContainerBuilder()
+            .build()
```
scholzj, line 62: This pattern seems to be a bit weird. Why aren't the options inside the builder but outside of it?

src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
```diff
@@ -77,11 +78,21 @@ public StrimziKafkaCluster(final int brokersNum, final int internalTopicReplicat
...
-                    .withNetworkAliases("broker-" + brokerId)
-                    .dependsOn(this.zookeeper)
-                    .withExternalZookeeper("zookeeper:" + StrimziZookeeperContainer.ZOOKEEPER_PORT);
+                StrimziKafkaContainer kafkaContainer = null;
+                try {
+                    kafkaContainer = new StrimziKafkaContainer.StrimziKafkaContainerBuilder()
+                        .withBrokerId(brokerId)
+                        .withKafkaConfigurationMap(additionalKafkaConfiguration)
+                        .withExternalZookeeperConnect("zookeeper:" + StrimziZookeeperContainer.ZOOKEEPER_PORT)
+                        .build()
```
scholzj, line 87: Same as above, seems a bit weird.

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -33,95 +33,81 @@
...
+
+        this.setDockerImageName("quay.io/strimzi-test-container/test-container:" +
+            this.strimziTestContainerVersion + "-kafka-" +
+            this.kafkaVersion);
+
+        LOGGER.info("=============================================");
+        LOGGER.info(this.getDockerImageName());
+        LOGGER.info("=============================================");
+
+        LOGGER.info("This is inside this object:{}", this::toString);
```
scholzj, line 87: Are these intentional? Or some left-over debug messages?

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -33,95 +33,81 @@
...
-     * @param brokerId broker id
-     * @return instance of StrimziKafkaContainer
+     * Default Kafka port
      */
-    public static StrimziKafkaContainer create(final int brokerId) {
-        return new StrimziKafkaContainer(brokerId, false, Collections.emptyMap());
-    }
+    public static final int KAFKA_PORT = 9092;
+    public static int kafkaDynamicKafkaPort;
+    private static final String DEFAUT_STORAGE_UUID = "xtzWWN4bTjitpL4efd9s6g";
```
scholzj, line 95: Is there some value in having this hardcoded? IIRC it was quite easy to generate it.

src/main/java/io/strimzi/utils/LogicalKafkaVersionEntity.java
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ *  A logical Kafka version
+ */
+public class LogicalKafkaVersionEntity {
+
+    private static final Logger LOGGER = LogManager.getLogger(LogicalKafkaVersionEntity.class);
+
+    private static final Pattern STRIMZI_TEST_CONTAINER_IMAGE_WITHOUT_KAFKA_VERSION = Pattern.compile("^test-container:(\\d+\\.\\d+\\.\\d+|latest)-kafka-.*$");
+    private static final String KAFKA_VERSIONS_URL_JSON = "https://raw.githubusercontent.com/strimzi/test-container-images/main/kafka_versions.yaml"; // this would be replaced by test-container-images url where json will be stored
```
scholzj, line 30: I forgot what we discussed before. Is this pulled every time the container is used? Do we have some _offline_ fallback?

src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
see-quick: It was because I did not also extend the Builder instance from `GenericContainer<StrimziKafkaContainerBuilder`. Now it should look like better :)).

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
see-quick: left-over removed...

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
see-quick: I have added a method in `Utils.randomUuid()` to generate such UUID.... (anyway this code was from Clement... :) )

src/main/java/io/strimzi/utils/LogicalKafkaVersionEntity.java
see-quick: If you run tests via `mvn` or `IDE` this is fetched only once but obviously when you run it multiple times it will fetch as many times as you execute it.

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -33,95 +34,76 @@
...
-     * @return instance of @code{StrimziKafkaContainer}
-     */
-    public static StrimziKafkaContainer createWithAdditionalConfiguration(final int brokerId, final Map<String, String> additionalKafkaConfiguration) {
-        return createWithAdditionalConfiguration(brokerId, false, additionalKafkaConfiguration);
+        // we need this shared network in case we deploy StrimziKafkaCluster which consist of `StrimziKafkaContainer`
+        // instances and by default each container has its own network, which results in `Unable to resolve address: zookeeper:2181`
+        this.withNetwork(Network.SHARED);
+        // exposing kafka port from the container
+        this.withExposedPorts(KAFKA_PORT);
+        this.withEnv("LOG_DIR", "/tmp");
```
scholzj, line 80: This is a bit weird since the `with*` methods are normally part of the builder pattern. Should these be just `set*` methods? It seems like we are mixing the things around.

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -33,95 +34,76 @@
...
     /**
-     * Static factory method, which creates @code{StrimziKafkaContainer} with empty additional configuration of Kafka broker
-     * @param brokerId broker id
-     * @return instance of StrimziKafkaContainer
+     * Default Kafka port
      */
-    public static StrimziKafkaContainer create(final int brokerId) {
-        return new StrimziKafkaContainer(brokerId, false, Collections.emptyMap());
-    }
+    public static final int KAFKA_PORT = 9092;
```
scholzj, line 90: Can we keep some order in the files? Class variables up top, then constructor, then other methods etc.?

src/main/java/io/strimzi/test/container/StrimziZookeeperContainer.java
```diff
@@ -24,35 +24,52 @@
...
+        this.withEnv("ZOOKEEPER_CLIENT_PORT", String.valueOf(ZOOKEEPER_PORT));
         // env for readiness
-        withEnv("ZOO_4LW_COMMANDS_WHITELIST", "ruok");
+        this.withEnv("ZOO_4LW_COMMANDS_WHITELIST", "ruok");
     }
 
+    /**
+     * Default ZooKeeper port
+     */
+    public static final int ZOOKEEPER_PORT = 2181;
```
scholzj, line 67: Sae as in the other class.

src/main/java/io/strimzi/utils/Utils.java
```diff
@@ -78,4 +80,14 @@ public static long waitFor(String description, long pollIntervalMs, long timeout
...
+
+    /**
+     * Static factory method to get a type 4 (pseudo randomly generated) UUID.
+     * @return random Uuid
+     */
+    public static Uuid randomUuid() {
+        UUID uuid = UUID.randomUUID();
+
+        return new Uuid(uuid.getMostSignificantBits(), uuid.getLeastSignificantBits());
+    }
```
scholzj, line 92: Does this work? IIRC Kafka had its own method for the Kraft storage ID. (as long as it does, we might not care)

src/main/java/io/strimzi/test/container/StrimziZookeeperContainer.java
see-quick: I have created a `Constants` for this one... :)

scholzj: It was more the ordering of class fields, methods etc. I meant here really - not the existence of this field in this class (I don't necessarily mind having a Constants class ... but that wasn't really how I meant it here.)

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
see-quick: I have changed it to set methods...(but those methods basically encapsulate with methods)

src/main/java/io/strimzi/utils/Utils.java
see-quick: This is basically a sub-set implementation of their method. Kafka uses:
```java
/**
     * Static factory to retrieve a type 4 (pseudo randomly generated) UUID.
     */
    public static Uuid randomUuid() {
        java.util.UUID uuid = java.util.UUID.randomUUID();
        while (uuid.equals(METADATA_TOPIC_ID_INTERNAL) || uuid.equals(ZERO_ID_INTERNAL)) {
            uuid = java.util.UUID.randomUUID();
        }
        return new Uuid(uuid.getMostSignificantBits(), uuid.getLeastSignificantBits());
    }`
```
and I think we do not need this part for topics:
```java
while (uuid.equals(METADATA_TOPIC_ID_INTERNAL) || uuid.equals(ZERO_ID_INTERNAL)) {
            uuid = java.util.UUID.randomUUID();
        }
```

scholzj: @tombentley Does this sound fine?

scholzj: I mean the part about us not needed the topic checks -> not sure if they are important for the cluster UUID.

see-quick: ^^ @tombentley

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -272,6 +201,66 @@ protected void containerIsStarting(final InspectContainerResponse containerInfo,
...
+    /**
+     * Fluent method, which sets @code{storageUUID}.
+     * The storage UUID must be the 16 bytes of a base64-encoded UUID.
+     *
+     * This method requires the broker to be created using Kraft.
+     *
+     * @param uuid the uuid, must not be {@code null} or blank.
+     * @return StrimziKafkaContainerBuilder instance
+     */
+    public StrimziKafkaContainer withStorageUUID(final String uuid) {
```
tombentley, line 256: What's the use case for allowing passing a uuid? How would be expect the caller to generate a string with the right syntax? Should we be using String? If so, should we validate it here?

src/main/java/io/strimzi/utils/LogicalKafkaVersionEntity.java
```diff
@@ -0,0 +1,192 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ *  A logical Kafka version
+ */
+public class LogicalKafkaVersionEntity {
```
tombentley, line 27: "Entity" doesn't really explain what's going on here. The inner class represents a concrete "Kafka version" and the outer class is more like a service that allows you to discover a concrete version that satisfies some more abstract criteria. So I would rename `LogicalKafkaVersionEntity` → `KafkaVersionService` and `LogicalKafkaVersion` → `KafkaVersion`. I think that naming would make it a bit clearer what's going on.

src/main/java/io/strimzi/utils/LogicalKafkaVersionEntity.java
```diff
@@ -0,0 +1,192 @@
...
+        public String toString() {
+            return "LogicalKafkaVersionEntity{" +
+                "version='" + this.version + '\'' +
+                ", image='" + this.image + '\'' +
+                '}';
+        }
+    }
+
+    /**
+     * Get the latest release
```
tombentley, line 114: Let's clarify what we mean by "latest" here, and explain how the result is intentionally not deterministic. It's the released `test-container-images` image with the highest Kafka version number. The result will change when a new version of Kafka is released with a higher Kafka version number, and a `test-container-images` image which contains that Kafka version has been built and released.

src/main/java/io/strimzi/utils/LogicalKafkaVersionEntity.java
```diff
@@ -0,0 +1,192 @@
...
+        return latestRelease;
+    }
+
+    /**
+     * Gets the previous minor release.
+     * If this instance were 2.8.1 then at a certain point in time this method
+     * might return a version reflecting 2.7.2.
+     * On a later date it might return 2.7.3.
+     * It is allowed to cross major version boundaries.
+     * E.g. the previous minor release to 3.0.0 might be 2.8.2
```
tombentley, line 142: This isn't really explaining to the caller when the result might change. I know it's difficult to explain. I think you should say that this is relative to `latestRelease()`. Illustrating with a table of latestRelease and previousMinor might help.

src/main/java/io/strimzi/utils/LogicalKafkaVersionEntity.java
```diff
@@ -0,0 +1,192 @@
...
+
+/**
+ *  A logical Kafka version
+ */
+public class LogicalKafkaVersionEntity {
+
+    private static final Logger LOGGER = LogManager.getLogger(LogicalKafkaVersionEntity.class);
+
+    private static final Pattern STRIMZI_TEST_CONTAINER_IMAGE_WITHOUT_KAFKA_VERSION = Pattern.compile("^test-container:(\\d+\\.\\d+\\.\\d+|latest)-kafka-.*$");
+    private static final String KAFKA_VERSIONS_URL_JSON = "https://raw.githubusercontent.com/strimzi/test-container-images/main/kafka_versions.yaml"; // this would be replaced by test-container-images url where json will be stored
```
tombentley, line 32: Shouldn't this point to a branch, rather than main?

Is there value (e.g. testing?) in allowing a system property to override the URL?

src/main/java/io/strimzi/utils/Utils.java
tombentley: `METADATA_TOPIC_ID_INTERNAL` and `ZERO_ID_INTERNAL` are two sentinel UUIDs. I don't think the `while` check is needed even in the Kafka code, since `java.util.UUID.randomUUID` returns a "type 4" UUID, which by definition has non-zero bytes in places where those sentinels have zero bytes.

tombentley (commented):
Mostly nits

src/main/java/io/strimzi/utils/LogicalKafkaVersionEntity.java
see-quick: I think it's enough to point the `main` explicitly and when we make a release in `test-container-images` in  `release-0.26.x` branch we change to `0.26.x`.

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
see-quick: I am thinking about that if it's necessary to allow user to specify `storageUUID`...

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -32,120 +34,61 @@
...
-     */
-    public static final int KAFKA_PORT = 9092;
-    /**
-     * Default ZooKeeper port
-     */
-    public static final int ZOOKEEPER_PORT = 2181;
-
     private static final String STARTER_SCRIPT = "/testcontainers_start.sh";
+    private static final KafkaVersionService LOGICAL_KAFKA_VERSION_ENTITY;
+    private static int kafkaDynamicKafkaPort;
```
tombentley, line 41: No need for two Kafkas surely, `dynamicKafkaPort` or `kafkaDynamicPort`?

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -32,120 +34,61 @@
...
-     * @param brokerId broker id
-     * @return instance of StrimziKafkaContainer
-     */
-    public static StrimziKafkaContainer createWithKraft(final int brokerId) {
-        return new StrimziKafkaContainer(brokerId, true, Collections.emptyMap());
+    private void buildDefaults() {
+        String strimziTestContainerVersion;
+        if (this.strimziTestContainerVersion == null || this.strimziTestContainerVersion.isEmpty()) {
+            strimziTestContainerVersion = LOGICAL_KAFKA_VERSION_ENTITY.latestRelease().getStrimziTestContainerVersion();
+            LOGGER.info("You did not specify Strimzi test container version. Using latest release:{}", strimziTestContainerVersion);
```
tombentley, line 60: I think using You in log messages is a bit unconventional: "No Strimzi test container version specified" would be better.

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -32,120 +34,61 @@
...
+            kafkaVersion = LOGICAL_KAFKA_VERSION_ENTITY.latestRelease().getVersion();
+            LOGGER.info("You did not specify Kafka version. Using latest release:{}", kafkaVersion);
+        } else {
+            kafkaVersion = this.kafkaVersion;
+        }
+        this.storageUUID = this.storageUUID == null ? Utils.randomUuid().toString() : this.storageUUID;
+
+        // we need this shared network in case we deploy StrimziKafkaCluster which consist of `StrimziKafkaContainer`
+        // instances and by default each container has its own network, which results in `Unable to resolve address: zookeeper:2181`
+        super.setNetwork(Network.SHARED);
```
tombentley, line 75: Any reason for the explicit `super`?

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -272,6 +201,99 @@ protected void containerIsStarting(final InspectContainerResponse containerInfo,
...
+     * Fluent method, which sets @code{externalZookeeperConnect}.
+     *
+     * If the broker was created using Kraft, this method throws an {@link IllegalArgumentException}.
+     *
+     * @param externalZookeeperConnect connect string
+     * @return StrimziKafkaContainer instance
+     */
+    public StrimziKafkaContainer withExternalZookeeperConnect(final String externalZookeeperConnect) {
+        if (useKraft) {
+            throw new IllegalArgumentException("Cannot configure an external Zookeeper and use Kraft at the same time");
```
tombentley, line 228: IllegalStateException would be better. The problem is that this method is being called _at all_, rather than there being a problem with the argument with which it is being called.

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
tombentley: Did you come to a conclusion?

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -272,6 +201,99 @@ protected void containerIsStarting(final InspectContainerResponse containerInfo,
...
+     * The storage UUID must be the 16 bytes of a base64-encoded UUID.
+     *
+     * This method requires the broker to be created using Kraft.
+     *
+     * @param uuid the uuid, must not be {@code null} or blank.
+     * @return StrimziKafkaContainerBuilder instance
+     */
+    public StrimziKafkaContainer withStorageUUID(final String uuid) {
+        if (!useKraft) {
+            throw new IllegalArgumentException("Setting the storage UUID requires Kraft");
```
tombentley, line 291: Again, for me ISE would be more appropriate here.

src/main/java/io/strimzi/utils/KafkaVersionService.java
```diff
@@ -0,0 +1,237 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ *  A logical Kafka version
```
tombentley, line 25:
```suggestion
 *  A service for querying for Kafka versions using abstract criteria such as "latest version", the result of which may change over time (and thus, non-deterministically).
```

tombentley (approved):
There are a few nits.

Does the `README.md` need to be updated to show how you can spin up a broker with a logical version?

Should `KafkaVersionService` support looking up a given version from a String, so that it's possible to write deterministic tests (example use case: I had to work around some issue in Kafka `x.y.z` that's fixed in `x.y.(z+1)`, and I want to write a regression test). It might make sense to do this in another PR, so we can merge this one.

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
see-quick: I think it does not matter whether I use explicit `super` or `this`. Because I did not override that method I am just using it for clarity that this method is from the `parent`.

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
see-quick: @cescoffier implement it so I assume they are using it...Thus I think we should implement also some validation mechanism, which can be copy pasted from Kafka source:
```java
if (str.length() > 24) {
            throw new IllegalArgumentException("Input string with prefix `"
                + str.substring(0, 24) + "` is too long to be decoded as a base64 UUID");
        }

        ByteBuffer uuidBytes = ByteBuffer.wrap(Base64.getUrlDecoder().decode(str));
        if (uuidBytes.remaining() != 16) {
            throw new IllegalArgumentException("Input string `" + str + "` decoded as "
                + uuidBytes.remaining() + " bytes, which is not equal to the expected 16 bytes "
                + "of a base64-encoded UUID");
        }
```

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
see-quick: Och, we need it :)) moving it to instance variable... )

see-quick:
> Should KafkaVersionService support looking up a given version from a String, so that it's possible to write deterministic tests (example use case: I had to work around some issue in Kafka x.y.z that's fixed in x.y.(z+1), and I want to write a regression test). It might make sense to do this in another PR, so we can merge this one.

Noted for the next PR :)

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
cescoffier: There is no real use case. At some point I would like to try to have the directory structure already created in the image. 

We can definitely remove that option and use the "default" value I provided (or any other default value)

src/test/java/io/strimzi/test/container/StrimziKafkaContainerIT.java
```diff
@@ -29,7 +29,8 @@ private void assumeDocker() {
     @Test
     void testStartContainerWithEmptyConfiguration() {
         assumeDocker();
-        systemUnderTest = StrimziKafkaContainer.create(1);
+        systemUnderTest = new StrimziKafkaContainer()
+            .withBrokerId(1);
```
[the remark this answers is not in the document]
see-quick: Good point done )

src/main/java/io/strimzi/utils/KafkaVersionService.java
```diff
@@ -0,0 +1,238 @@
...
+    private final List<KafkaVersion> logicalKafkaVersionEntities = new ArrayList<>();
+
+    /**
+     * Constructor of the KafkaVersionService, which invokes resolution and parsing phase. In resolution, we fetch
+     * data from @code{KAFKA_VERSIONS_URL_JSON} and then we parse that json scheme, which gives us list of @code{KafkaVersion}
+     * objects.
+     */
+    public KafkaVersionService() {
+        // scrape json schema and fill the inner list of versions
+        this.resolveAndParse();
```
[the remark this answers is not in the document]
see-quick: > I often work offline (yes, I'm old school)

Just turn off your notification :D. Any way I can implement some cache mechanisms... ))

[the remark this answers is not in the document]
see-quick: >  I often work offline (yes, I'm old school) to avoid notifications. How would that work in this case?

By current implementation, it will not work.

src/main/java/io/strimzi/utils/KafkaVersionService.java
```diff
@@ -0,0 +1,238 @@
...
+/**
+ *  A service for querying for Kafka versions using abstract criteria such as "latest version",
+ *  the result of which may change over time (and thus, non-deterministically).
+ */
+public class KafkaVersionService {
+
+    private static final Logger LOGGER = LogManager.getLogger(KafkaVersionService.class);
+
+    private static final Pattern STRIMZI_TEST_CONTAINER_IMAGE_WITHOUT_KAFKA_VERSION = Pattern.compile("^test-container:(\\d+\\.\\d+\\.\\d+|latest)-kafka-.*$");
+    private static final String KAFKA_VERSIONS_URL_JSON = "https://raw.githubusercontent.com/strimzi/test-container-images/main/kafka_versions.yaml"; // this would be replaced by test-container-images url where json will be stored
```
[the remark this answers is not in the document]
see-quick: Changed to json )

pom.xml
```diff
@@ -240,7 +252,7 @@
                         </goals>
                         <configuration>
                             <compilerArgs>
-                                <arg>-Xlint:unchecked,deprecation</arg>
+                                <arg>-Xlint:unchecked</arg>
```
scholzj, line 255: Why remove deprecation?

src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
```diff
@@ -77,11 +77,15 @@ public StrimziKafkaCluster(final int brokersNum, final int internalTopicReplicat
...
+                    .withBrokerId(brokerId)
+                    .withKafkaConfigurationMap(additionalKafkaConfiguration)
+                    .withExternalZookeeperConnect("zookeeper:" + Constants.ZOOKEEPER_PORT)
                     .withNetwork(this.network)
                     .withNetworkAliases("broker-" + brokerId)
-                    .dependsOn(this.zookeeper)
-                    .withExternalZookeeper("zookeeper:" + StrimziZookeeperContainer.ZOOKEEPER_PORT);
+                    .dependsOn(this.zookeeper);
+
+                LOGGER.info("Started broker with id:{}", kafkaContainer);
```
scholzj, line 88: Space after `:`?

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -6,7 +6,9 @@
 
 import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.api.model.ContainerNetwork;
-import io.strimzi.utils.Environment;
+import io.strimzi.utils.Constants;
+import io.strimzi.utils.KafkaVersionService;
+import io.strimzi.utils.Utils;
```
scholzj, line 11: Should this be something like `io.strimzi.test.container.utils`?

src/main/java/io/strimzi/utils/KafkaVersionService.java
```diff
@@ -0,0 +1,238 @@
...
+     * | (prev prev minor) ---  (previous minor) ---  (current) |
+     *                                                     |
+     *                                                     *
+     * |    2.8.1   &lt;---&gt;     2.8.2  &lt;---&gt;   3.0.0  |
+     * ============================================================
+     * Assuming that test container images `kafka_versions.yaml` has following content:
+     * {
+     *   "version": 1,
+     *   "kafkaVersions": {
+     *      2.8.1": "test-container:latest-kafka-2.8.1",
```
scholzj, line 184:
```suggestion
     *      "2.8.1": "test-container:latest-kafka-2.8.1",
```

src/main/java/io/strimzi/utils/KafkaVersionService.java
```diff
@@ -0,0 +1,238 @@
...
+            throw new IllegalStateException("Wrong json schema! It must have at least one release");
+        }
+
+        // sort in this case always order the latest release at the end so the previous one is the previous minor release
+        this.logicalKafkaVersionEntities.sort(KafkaVersion::compareTo);
+
+        // 1.0.0
+        // 2.40.50
+        // 3.0.0 <- previous minor (it will be always previous last)
+        // 4.1.2 <- latest release
```
scholzj, line 202: Is this some leftover which should be deleted?

pom.xml
see-quick: Huh, thanks for pointing it out. This is another problem we have to solve. Test container from `1.5.x` version deprecated standard `GenericContainer` constructor. In our case, it's `StrimziKafkaContainer` and `StrimziZookeeperContainer` and the possible re-placement is to use a constructor where the user has to specify `DockerImageName`, which is what we want to avoid... ))

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
see-quick: No these are auxiliary methods used in the src code

src/main/java/io/strimzi/utils/KafkaVersionService.java
see-quick: we can remove it :) thanks

pom.xml
scholzj: Well, the question is whether disabling it here is the right approach. I think you should keep it here enabled and instead use annotation to suppress the deprecation warning where needed. That is what we follow in operators etc.

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
scholzj: Yeah, I understand it. But I think having a clear separation at the package level would be good. I.e. having all classes here in the test container package.

pom.xml
see-quick: Make sense :) Will do

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
see-quick: Okay :)

scholzj (approved):

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -30,122 +32,62 @@
  * The additional configuration for Kafka broker can be injected via constructor. This container is a good fit for
  * integration testing but for more hardcore testing we suggest using @StrimziKafkaCluster.
  */
+@SuppressWarnings("deprecation")
```
scholzj, line 35: You should have a comment about why is it here - to remember what was deprecated to remove this.

PR #8, opened 2021-11-20
Extend the StrimziKafkaContainer to allow starting Kafka with KRaft.

cescoffier:
The goal of this PR is to enable Kraft and reduce startup time. With it, I'm around 4-5 seconds to start the Kafka container. With such startup time, this Test Container can be used as the Quarkus Dev Services for Kafka. It also allows projects using Kafka to have an easy way to verify their behavior when KRaft is used. 

This PR: 

* Adds support for Kraft in the StrimziKafkaContainer
* Allows configuring the storage UUID (advanced option)
* Updates scripts to support Kraft (changed override and avoid starting Zookeeper)
* Extends test suite with specific Kraft tests


src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -156,13 +207,22 @@ protected void containerIsStarting(final InspectContainerResponse containerInfo,
...
-            .append(" --override advertised.listeners=").append(advertisedListeners)
-            .append(" --override zookeeper.connect=localhost:").append(ZOOKEEPER_PORT)
-            .append(" --override listener.security.protocol.map=").append(kafkaListenerSecurityProtocol).append("PLAINTEXT:PLAINTEXT")
-            .append(" --override inter.broker.listener.name=BROKER1");
+        if (!useKraft) {
+            // default listener config
+            kafkaConfiguration
+                    .append(" --override listeners=").append(kafkaListeners).append("PLAINTEXT://0.0.0.0:").append(KAFKA_PORT)
+                    .append(" --override advertised.listeners=").append(advertisedListeners)
+                    .append(" --override zookeeper.connect=localhost:").append(ZOOKEEPER_PORT)
```
see-quick, line 215: As I can see the only thing that differs from using KRaft is `.append(" --override zookeeper.connect=localhost:").append(ZOOKEEPER_PORT)`. Would It be better to make 
```
kafkaConfiguration
                    .append(" --override listeners=").append(kafkaListeners).append("PLAINTEXT://0.0.0.0:").append(KAFKA_PORT)
                    .append(" --override advertised.listeners=").append(advertisedListeners)
                    .append(" --override listener.security.protocol.map=").append(kafkaListenerSecurityProtocol).append("PLAINTEXT:PLAINTEXT")
                    .append(" --override controller.listener.names=").append("BROKER1")
                    .append(" --override inter.broker.listener.name=BROKER1");
```
and afterwards use decision logic to only append `.append(" --override zookeeper.connect=localhost:").append(ZOOKEEPER_PORT)` if we don't use Kraft?

see-quick (commented):
First of all thanks for such contribution @cescoffier. Only one nit...

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
tombentley: There's also `.append(" --override controller.listener.names=").append("BROKER1")` in the Kraft case, but I agree we should try to keep the common configs out of the `if/else`.

src/test/java/io/strimzi/test/container/StrimziKafkaKraftContainerIT.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private void assumeDocker() {
+        Assumptions.assumeTrue(System.getenv("DOCKER_CMD") == null || "docker".equals(System.getenv("DOCKER_CMD")));
+    }
+
+    @Test
+    void testStartContainerWithEmptyConfiguration() throws ExecutionException, InterruptedException {
+        assumeDocker();
+        systemUnderTest = StrimziKafkaContainer.createWithKraft(1)
+            .waitingFor(Wait.forLogMessage(".*Transitioning from RECOVERY to RUNNING.*", 1));
```
tombentley, line 46: Shouldn't we encapsulate this via some `waitForRunning` method? It would save users having to know the regex, and we could insulate them should the log message ever change.

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
cescoffier: Fixed.

cescoffier:
@see-quick @tombentley Thanks for your comments! I've pushed an updated version:

* mentions KIP-500
* adds `waitForRunning`
* extracts the common configuration outside of the `if(kraft)` blocks

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -104,15 +131,54 @@ protected void doStart() {
...
 
+    /**
+     * Fluent method, which sets a waiting strategy to wait until the broker is ready.
+     *
+     * This method waits for a log message in the broker log.
+     * You can customize the strategy using {@link #waitingFor(WaitStrategy)}.
+     *
+     * @return StrimziKafkaContainer instance
+     */
+    public StrimziKafkaContainer withForRunning() {
```
tombentley, line 157: I guess this should be `waitForRunning`, or possibly `withWaitForRunning`.

tombentley (approved):
Just a method name nit, but otherwise LGTM. Thanks!

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
cescoffier: Oh.... I write the right version in the README and not in the code. Let me fix that (`waitForRunning()` it is.)

cescoffier: Fixed.

see-quick (approved):

see-quick:
Thanks for the PR :)

PR #18, opened 2022-01-18
[MO] - adding kafka version json and system property

see-quick:
This PR changes the logic of how we fetch supported versions of Kafka with supported images:

- removing dependency on the test-container-images repository (kafka_versions.json - now it's embedded in this repo)
- introducing a new Java system property by which we can specify custom images `strimzi.custom.image`
- update `Readme.md`


src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -68,7 +66,9 @@ public StrimziKafkaContainer() {
 
     @Override
     protected void doStart() {
-        this.setDockerImageName(KafkaVersionService.strimziTestContainerImageName(strimziBaseImage, strimziTestContainerImageVersion, kafkaVersion));
+        System.out.println("What is this");
```
scholzj, line 69: I guess this should be deleted? :-o

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -68,7 +66,9 @@ public StrimziKafkaContainer() {
...
+        System.out.println(KafkaVersionService.strimziTestContainerImageName(kafkaVersion));
```
scholzj, line 70: So should this?

src/main/java/io/strimzi/test/container/utils/KafkaVersionService.java
```diff
@@ -28,11 +23,7 @@
 public class KafkaVersionService {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(KafkaVersionService.class);
-
-    private static final Pattern STRIMZI_TEST_CONTAINER_IMAGE_WITHOUT_KAFKA_VERSION = Pattern.compile("^test-container:(\\d+\\.\\d+\\.\\d+|latest)-kafka-.*$");
-    private static final String KAFKA_VERSIONS_URL_JSON = "https://raw.githubusercontent.com/strimzi/test-container-images/main/kafka_versions.json";
-    private static final String STRIMZI_BASE_IMAGE = "quay.io/strimzi-test-container/test-container";
-    private static final String IMAGE_FORMAT = "%s:%s-kafka-%s";
+    private static final Properties PROPERTIES = System.getProperties();
```
scholzj, line 26: Any special reason why to store this instead of just calling `System.getProperties()` in the place where you use it?

src/main/java/io/strimzi/test/container/utils/KafkaVersionService.java
```diff
@@ -61,29 +52,37 @@ public static KafkaVersionService getInstance() {
...
-        if (kafkaVersion == null || kafkaVersion.isEmpty()) {
-            kafkaVersion = KafkaVersionService.getInstance().latestRelease().getVersion();
-            LOGGER.info("No Kafka version specified. Using latest release:{}", kafkaVersion);
+    public static String strimziTestContainerImageName(String kafkaVersion) {
+        String imageName = null;
+        final Object strimziCustomImageName = PROPERTIES.get("strimzi.custom.image");
+
+        if (strimziCustomImageName != null && !strimziCustomImageName.toString().isEmpty()) {
+            final String customImage = strimziCustomImageName.toString();
+            LOGGER.info("Using custom image:{}", customImage);
```
scholzj, line 68: Space after `:`?

src/main/java/io/strimzi/test/container/utils/KafkaVersionService.java
```diff
@@ -61,29 +52,37 @@ public static KafkaVersionService getInstance() {
...
+            // using custom image provided from SystemProperty
+            imageName = customImage;
+        } else {
+            // using strimzi-test-container images
+            if (kafkaVersion == null || kafkaVersion.isEmpty()) {
+                imageName = KafkaVersionService.getInstance().latestRelease().getImage();
+                kafkaVersion = KafkaVersionService.getInstance().latestRelease().getImage();
+                LOGGER.info("No Kafka version specified. Using latest release:{}", kafkaVersion);
```
scholzj, line 76: Space after `:`?

src/main/java/io/strimzi/test/container/utils/KafkaVersionService.java
```diff
@@ -248,10 +228,7 @@ public KafkaVersion previousMinor() {
     private void resolveAndParse() {
         // Connect to the URL using java's native library
```
scholzj, line 229: The comment does not apply anymore I guess?

src/test/java/io/strimzi/test/container/StrimziKafkaContainerIT.java
```diff
@@ -125,8 +125,10 @@ void testStartContainerWithServerProperties() {
     void testStartContainerWithStrimziKafkaImage() {
         assumeDocker();
 
+        // explicitly set strimzi.custom.image
+        System.setProperty("strimzi.custom.image", "quay.io/strimzi/kafka:latest-kafka-3.0.0");
```
scholzj, line 129: Should this be `quay.io/strimzi-test-container/test-container:latest-kafka-3.0.0`? Or if it is just a dummy value to see it is passed properly, it should be something dummy to make it clear what value does it have.

see-quick: In this test, we actually spin up Kafka container so it's needed :).

scholzj: So, how do you know it is/will be compatible? Should you at least use a release version which will not change anymore?

scholzj:
PS: I guess the Lift issues can be somehow fixed as well? Should it be `{@code ...}`?

src/test/java/io/strimzi/test/container/StrimziKafkaContainerIT.java
see-quick: > Should you at least use a release version that will not change anymore?

Will do :)

src/main/java/io/strimzi/test/container/StrimziZookeeperContainer.java
```diff
@@ -54,7 +52,7 @@ public StrimziZookeeperContainer() {
 
     @Override
     protected void doStart() {
-        this.setDockerImageName(KafkaVersionService.strimziTestContainerImageName(strimziBaseImage, strimziTestContainerImageVersion, kafkaVersion));
+        this.setDockerImageName(KafkaVersionService.strimziTestContainerImageName(kafkaVersion));
         // we need it for the startZookeeper(); and startKafka(); to run container before...
```
tombentley, line 56: I don't understand this comment. What is "it"?

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -68,7 +66,7 @@ public StrimziKafkaContainer() {
 
     @Override
     protected void doStart() {
-        this.setDockerImageName(KafkaVersionService.strimziTestContainerImageName(strimziBaseImage, strimziTestContainerImageVersion, kafkaVersion));
+        this.setDockerImageName(KafkaVersionService.strimziTestContainerImageName(kafkaVersion));
         // we need it for the startZookeeper(); and startKafka(); to run container before...
```
tombentley, line 70: I don't understand this comment. What is "it"?

src/main/java/io/strimzi/test/container/utils/KafkaVersionService.java
```diff
@@ -61,29 +50,37 @@ public static KafkaVersionService getInstance() {
...
-        }
-        if (strimziTestContainerImageVersion == null || strimziTestContainerImageVersion.isEmpty()) {
-            strimziTestContainerImageVersion = KafkaVersionService.getInstance().latestRelease().getStrimziTestContainerVersion();
-            LOGGER.info("No Strimzi test container version specified. Using latest release:{}", strimziTestContainerImageVersion);
-        }
-
-        if (kafkaVersion == null || kafkaVersion.isEmpty()) {
-            kafkaVersion = KafkaVersionService.getInstance().latestRelease().getVersion();
-            LOGGER.info("No Kafka version specified. Using latest release:{}", kafkaVersion);
+    public static String strimziTestContainerImageName(String kafkaVersion) {
```
tombentley, line 60: I'm not sure this really needs to be public? I know you have this `utils` package, which is why you need it to be public at the moment, but it's not clear to me why you really need a separate package. Having a single package would mean you could make this (and probably other methods) package access, which would reduce the API surface which you're exposing to uses, which in turn makes future refactoring much easier to reason about (they literally couldn't be calling internal methods like this one).

README.md
```diff
@@ -143,26 +143,37 @@ StrimziKafkaContainer strimziKafkaContainer = new StrimziKafkaContainer()
...
+If kafka version is not set then the latest version is configured automatically.
+
+#### viii) (Optional) Specify custom image
+
+In case you want to use your custom image (i.e., not from `src/main/java/resources/kafka_versions.json`) and 
+use for instance Strimzi base image you can achieve it via System property `strimzi.custom.image`.
+
+```java
+// explicitly set strimzi.custom.image
+System.setProperty("strimzi.custom.image", "quay.io/strimzi/kafka:0.27.1-kafka-3.0.0");
```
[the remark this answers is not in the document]
see-quick: yes make sense

src/main/java/io/strimzi/test/container/StrimziZookeeperContainer.java
```diff
@@ -54,7 +52,7 @@ public StrimziZookeeperContainer() {
 
     @Override
     protected void doStart() {
-        this.setDockerImageName(KafkaVersionService.strimziTestContainerImageName(strimziBaseImage, strimziTestContainerImageVersion, kafkaVersion));
+        this.setDockerImageName(KafkaVersionService.strimziTestContainerImageName(kafkaVersion));
```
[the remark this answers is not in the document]
see-quick: io.strimzi.test.container.utils.UnknownKafkaVersionException: Doesn't know the specified Kafka version: 2.4.0. The supported Kafka versions are: [2.8.1, 3.0.0] something like that?

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -68,7 +66,7 @@ public StrimziKafkaContainer() {
 
     @Override
     protected void doStart() {
-        this.setDockerImageName(KafkaVersionService.strimziTestContainerImageName(strimziBaseImage, strimziTestContainerImageVersion, kafkaVersion));
+        this.setDockerImageName(KafkaVersionService.strimziTestContainerImageName(kafkaVersion));
```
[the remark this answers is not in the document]
see-quick: Good I have added an exception and also write new test case for that.

src/main/java/io/strimzi/test/container/StrimziZookeeperContainer.java
see-quick: This command :)  
`sh", "-c", "while [ ! -f " + STARTER_SCRIPT + " ]; do sleep 0.1; done; " + STARTER_SCRIPT`

src/main/java/io/strimzi/test/container/utils/KafkaVersionService.java
see-quick: So should I remove the `utils` package?

src/main/java/io/strimzi/test/container/KafkaVersionService.java
```diff
@@ -61,35 +50,48 @@ public static KafkaVersionService getInstance() {
...
-            strimziTestContainerImageVersion = KafkaVersionService.getInstance().latestRelease().getStrimziTestContainerVersion();
-            LOGGER.info("No Strimzi test container version specified. Using latest release:{}", strimziTestContainerImageVersion);
-        }
-
-        if (kafkaVersion == null || kafkaVersion.isEmpty()) {
-            kafkaVersion = KafkaVersionService.getInstance().latestRelease().getVersion();
-            LOGGER.info("No Kafka version specified. Using latest release:{}", kafkaVersion);
+    protected static String strimziTestContainerImageName(String kafkaVersion) {
+        final String imageName;
+        final Object strimziCustomImageName = System.getProperties().get("kafka.custom.image");
```
scholzj, line 64: You can probably define it somewhere as a constant. I also wonder little bit if the name is the right nme. Should it have some Strimzi prefix? `strimzi.test-container.custom.image` or something like that to make it more distinguishable?

see-quick: @tombentley What do you think about this? :)

src/main/resources/kafka_versions.json
```diff
@@ -0,0 +1,6 @@
+{
+  "kafkaVersions": {
+    "2.8.1": "quay.io/strimzi-test-container/test-container:latest-kafka-2.8.1",
+    "3.0.0": "quay.io/strimzi-test-container/test-container:latest-kafka-3.0.0"
```
tombentley, line 4: I really feel we should consider embedding a version number for the contract between the container and `test-containers` in the tag or name of the images. We can't guarantee that we won't make changes in `test-container-images` in the future which are incompatible with the version of `test-container` which we release now. Using a tag like `v1-kafka-3.0.0` allows us to have a `v2-kafka-3.0.0` in the future and it not break existing uses of `test-container`. We may never need it, but having it allows us to to more easily change our minds.

scholzj: I think this is the master version. The release versions would not have `latest` but some version of the images in it.

see-quick: Yes as Jakub mentioned the release version of `test-container-images` would look like this:
```
 "3.0.0": "quay.io/strimzi-test-container/test-container:0.100.x-kafka-3.0.0"
```

So it should be okay or @tombentley?

tombentley (approved):

src/main/java/io/strimzi/test/container/KafkaVersionService.java
see-quick: Okay I will replaced it with `strimzi.test-container.kafka.custom.image` Thanks guys :)

PR #24, opened 2022-01-25
[MO] - adds ParametrizedTest for test cases

see-quick:
This PR adds ParametrizedTests for our test cases. Currently, we have been running only tests for the latest version. Now we would be able to catch errors on different supported Kafka versions. 

### DIFF
--- a/src/test/java/io/strimzi/test/container/AbstractIT.java
+++ b/src/test/java/io/strimzi/test/container/AbstractIT.java
@@ -43,7 +43,7 @@ public class AbstractIT {
         Assumptions.assumeTrue(System.getenv("DOCKER_CMD") == null || "docker".equals(System.getenv("DOCKER_CMD")));
     }
 
-    protected void assumeKafkaThirdVersionAndMore(final String imageName) {
+    protected void supportsKraftMode(final String imageName) {
         Assumptions.assumeTrue(!imageName.contains("-kafka-2."));
     }
 }
--- a/src/test/java/io/strimzi/test/container/AbstractIT.java
+++ b/src/test/java/io/strimzi/test/container/AbstractIT.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.test.container;
+
+import java.io.IOException;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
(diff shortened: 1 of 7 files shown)


Frawless (approved):

src/test/java/io/strimzi/test/container/AbstractIT.java
```diff
@@ -0,0 +1,49 @@
...
+            parameters.add(Arguments.of(fields.getValue().asText()));
+        }
+        return parameters.stream();
+    }
+
+    protected void assumeDocker() {
+        Assumptions.assumeTrue(System.getenv("DOCKER_CMD") == null || "docker".equals(System.getenv("DOCKER_CMD")));
+    }
+
+    protected void assumeKafkaThirdVersionAndMore(final String imageName) {
```
scholzj, line 46: I think this is confusing. If anything, it should be named `assumeKafkaThree` I guess? But maybe you should instead name it after the feature it is gating. E.g. `supportsKraftMode`.

src/test/java/io/strimzi/test/container/StrimziKafkaContainerIT.java
```diff
@@ -20,20 +21,17 @@
...
     private StrimziKafkaContainer systemUnderTest;
 
-    private void assumeDocker() {
-        Assumptions.assumeTrue(System.getenv("DOCKER_CMD") == null || "docker".equals(System.getenv("DOCKER_CMD")));
-    }
-
-    @Test
-    void testStartContainerWithEmptyConfiguration() {
+    @ParameterizedTest(name = "testStartContainerWithEmptyConfiguration-{0}")
+    @MethodSource("retrieveKafkaVersionsFile")
```
scholzj, line 31: Can't this be set on the class level?

scholzj (approved):
Some nits. But LGTM assuming it works.

src/test/java/io/strimzi/test/container/StrimziKafkaContainerIT.java
see-quick: I asked the same question to myself but this `@MethodSource` can be only set in method scope. :( Same applies to `@ArgumentsSource` and other source types, which JUnit5 provides... 

Another approach would be using `@TestFactory` where we could eliminate multiple calls of the retrieveKafkaVersionsFile method...but I don't know if that's worth it...because it would make tests harder to understand :)

scholzj: Well, I guess you can keep this then.

PR #25, opened 2022-01-25
[MO] - restrict KRaft for Kafka 2.8.1

see-quick:
This PR restrict usage of KRaft for Kafka `2.8.1` and it's allowed only for `3.x.x` Kafka versions. Moreover, I have added tests cases for sanity checks. 

### DIFF
--- a/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
+++ b/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
@@ -102,6 +102,7 @@ public class StrimziKafkaContainer extends GenericContainer<StrimziKafkaContaine
             }
         } catch (InterruptedException | ExecutionException e) {
             LOGGER.error("Error occurred during retrieving of image name provider", e);
+            throw new RuntimeException(e);
         }
         // we need it for the startZookeeper(); and startKafka(); to run container before...
         super.setCommand("sh", "-c", "while [ ! -f " + STARTER_SCRIPT + " ]; do sleep 0.1; done; " + STARTER_SCRIPT);
--- a/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
+++ b/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.function.Function;
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -92,8 +93,15 @@ private StrimziKafkaContainer(CompletableFuture<String> imageName) {
...
-            imageNameProvider.complete(KafkaVersionService.strimziTestContainerImageName(kafkaVersion));
+        if (!this.imageNameProvider.isDone()) {
+            this.imageNameProvider.complete(KafkaVersionService.strimziTestContainerImageName(this.kafkaVersion));
+        }
+        try {
+            if (this.useKraft && ((this.kafkaVersion != null && this.kafkaVersion.equals("2.8.1")) || this.imageNameProvider.get().contains("2.8.1"))) {
+                throw new UnsupportedKraftKafkaVersionException("Specified Kafka version " + this.kafkaVersion + " is not supported in KRaft mode.");
+            }
+        } catch (InterruptedException | ExecutionException e) {
+            LOGGER.error("Error occurred during retrieving of image name provider", e);
```
scholzj, line 104: I guess this should throw some other exception?

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -92,8 +93,15 @@ private StrimziKafkaContainer(CompletableFuture<String> imageName) {
 
     @Override
     protected void doStart() {
-        if (!imageNameProvider.isDone()) {
-            imageNameProvider.complete(KafkaVersionService.strimziTestContainerImageName(kafkaVersion));
+        if (!this.imageNameProvider.isDone()) {
+            this.imageNameProvider.complete(KafkaVersionService.strimziTestContainerImageName(this.kafkaVersion));
+        }
```
scholzj, line 98: Why is this first? I would expect the check which does nothing and just throws exception to be first before calling some other methods. Or is it this method which sets the fields used below?

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
see-quick: This exception is possibly thrown when I invoke `this.imageNameProvider.get().` That's why it is needed here

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
see-quick: >  Or is it this method which sets the fields used below?

Yes :)

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
scholzj: Sure, but why you just log it and ignore it? It clearly suggests the check didn't complete and there might be some issue.

see-quick: Ah, I see your point. I can add line, which will printStackTrace (`e.printStackTrace();`)

scholzj (approved):

PR #26, opened 2022-01-27
Controller listens on the localhost instead of container host

ozangunalp:
When container host IP is not on localhost kraft containers fail on startup.


see-quick (approved):

scholzj (approved):
LGTM assuming it works.

PR #32, opened 2022-03-01
Fix starter script test

cescoffier:
Extract the start command in its own method to allow consumers to override it if needed.
Also, fix the file test as the file must be executable (and not only existing).

Context: We are seeing a few "text file busy" when starting strimzi. It's related to a race condition on the file system as the file exists, but the executable flag is not set or not completely set. This PR waits for the file to be executable and allows overriding the command to let us find the best approach (if -x is not enough).


cescoffier:
@ozangunalp as discussed.

see-quick (approved):
Thanks for this @cescoffier. LGTM.

PR #37, opened 2022-05-17
Remove compile dependency on kafka-clients

TonyDemol:
The `kafka-clients` 3+ dependency is currently declared with `compile` scope, which makes the library harder to integrate in code bases when using a `kafka-clients` 2+ (code base where the migration to `kafka-client`` 3+ is not easily doable).

The dependency is only used to generate an UUID in the `StrimziKafkaContainer` class using the Kafka client's `org.apache.kafka.common.Uuid`, in order to apply a specific filter to ignore 2 specific Kafka UUIDs (metadata topic uuid & zero uuid) and to encode the UUID in base64.

The constraints to have a `compile` dependency on the `kafka-clients` seems very high compared to the benefits.

### DIFF
--- a/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
+++ b/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
@@ -6,7 +6,6 @@ package io.strimzi.test.container;
 
 import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.api.model.ContainerNetwork;
-import org.apache.kafka.common.Uuid;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.testcontainers.containers.GenericContainer;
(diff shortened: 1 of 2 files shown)


scholzj:
@tombentley WDYT about this? Is the UUID stable enough to replace it like this?

TonyDemol:
Hi! 
Is there any news about this PR?
Thanks in advance :)

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -280,6 +282,21 @@ private String writeOverrideString(Map<String, String> kafkaConfigurationMap) {
         return kafkaConfiguration.toString();
     }
 
+    private String randomUuid() {
```
tombentley, line 285: Please could you add a comment explaining that we're rolling our own uuid generation here, and replicating Kafka's logic to avoid the dependency.

tombentley (approved):

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
TonyDemol: Thanks for the approval and feedback. I added a comment. Let me know if this is what you were expecting (English is not my native tong, I hope this is understandable).

scholzj (approved):

scholzj:
Thanks for the PR.

TonyDemol:
Hi @scholzj,
Do you know if a date is planned for the 0.102.0 release?
Thanks in advance.

see-quick:
> Hi @scholzj, Do you know if a date is planned for the 0.102.0 release? Thanks in advance.

Yesterday, we released CR1 of the test container :) (https://github.com/strimzi/test-container/releases/tag/0.102.0-rc1)

TonyDemol:
Amazing :). Thanks a lot.

PR #42, opened 2022-06-28
Fix #40: Update test containers version

### LINKED ISSUE #40: Update testcontainers dependency
Currently we're using 1.15.2, but the latest upstream is 1.17.2.

tombentley:
* Update docker java version, since it's a transitive dep of testcontainers
* `getContainerIpAddress()` got deprecated, so use `getHost`()
* `systemUnderTest` can be null if the instantiation failed, so add a null check to avoid hiding the original exception.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,10 +89,10 @@
         <maven.compiler.source>1.8</maven.compiler.source>
 
         <!-- DEPENDENCY CODE VERSIONS       -->
-        <test-containers.version>1.15.2</test-containers.version>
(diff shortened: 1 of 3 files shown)


tombentley:
> Ot of curiosity ... what version of net.java.dev.jna:jna is this using? Is it already 5.8.0 or newer? I needed to update it to get it working on Arm64 based Macbook

This PR changes it from 5.2.0 to 5.8.0

scholzj:
Great, thanks.

scholzj (approved):

PR #44, opened 2022-07-07
Add proxy container configuration

fvaleri:
A proxy container (i.e. Toxiproxy) can be optionally configured.
This container allows to create a TCP proxy between test code and Kafka broker.

Every Kafka broker request will pass through the proxy where you can simulate
network conditions (i.e. connection cut, latency). For example, you can create
a network partition when running a multi-node cluster.

This commit also adds a common super-interface for containers.

Signed-off-by: Federico Valeri <fedevaleri@gmail.com>

### DIFF
--- a/.checkstyle/checkstyle.xml
+++ b/.checkstyle/checkstyle.xml
@@ -87,7 +87,10 @@
         <!-- code quality -->
         <module name="MethodLength"/>
         <module name="ParameterNumber"/>
-        <module name="ClassDataAbstractionCoupling"/>
+        <module name="ClassDataAbstractionCoupling">
+            <!-- default is 7 -->
+            <property name="max" value="10"/>
+        </module>
         <module name="BooleanExpressionComplexity"/>
         <module name="ClassFanOutComplexity"/>
         <module name="CyclomaticComplexity">
--- a/src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
+++ b/src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
@@ -8,7 +8,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.testcontainers.containers.Container;
 import org.testcontainers.containers.Network;
-import org.testcontainers.lifecycle.Startable;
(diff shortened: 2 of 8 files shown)


.checkstyle/checkstyle.xml
```diff
@@ -87,7 +87,10 @@
         <!-- code quality -->
         <module name="MethodLength"/>
         <module name="ParameterNumber"/>
-        <module name="ClassDataAbstractionCoupling"/>
+        <module name="ClassDataAbstractionCoupling">
+            <!-- default is 7 -->
+            <property name="max" value="10"/>
+        </module>
```
scholzj, line 93: Please use annotation rather than increasing the global limit.

scholzj (commented):
Maybe you can add to the description some more explanation as to how your changes related to the aim of the PR and why are they needed for the _proxy_ support since it is not really clear to me how some of your changes relate to it.

It would be also good to explain how does the proxy impact the advertised hosts in Kafka since that is not clear either.

see-quick (approved):

src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
```diff
@@ -37,17 +37,25 @@ public class StrimziKafkaCluster implements Startable {
...
+     * The {@code proxyContainer} allows to simulate network conditions (i.e. connection cut, latency).
+     * For example, you can simulate a network partition by cutting the connection of one or more brokers.
+     *
      * @param brokersNum number of brokers
      * @param internalTopicReplicationFactor internal topics
      * @param additionalKafkaConfiguration additional Kafka configuration
+     * @param proxyContainer Proxy container
      */
-    public StrimziKafkaCluster(final int brokersNum, final int internalTopicReplicationFactor, final Map<String, String> additionalKafkaConfiguration) {
+    public StrimziKafkaCluster(final int brokersNum,
```
scholzj, line 55: I guess this is a public interface which we are changing? Should we instead have another one for backwards compatibility?

fvaleri: Yes, good point.

PR #53, opened 2023-02-27
Update to the newest Kafka + some minor changes

see-quick:
This PR adds support for Kafka versions, which were introduced (https://github.com/strimzi/test-container-images/pull/14). Also, it adds some minor changes (i.e., changing regex for waiting until the cluster is ready).

### DIFF
--- a/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
+++ b/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
@@ -116,7 +116,7 @@ public class StrimziKafkaContainer extends GenericContainer<StrimziKafkaContaine
             this.imageNameProvider.complete(KafkaVersionService.strimziTestContainerImageName(this.kafkaVersion));
         }
         try {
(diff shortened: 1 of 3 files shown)


src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -116,7 +116,7 @@ protected void doStart() {
             this.imageNameProvider.complete(KafkaVersionService.strimziTestContainerImageName(this.kafkaVersion));
         }
         try {
-            if (this.useKraft && ((this.kafkaVersion != null && this.kafkaVersion.equals("2.8.1")) || this.imageNameProvider.get().contains("2.8.1"))) {
+            if (this.useKraft && ((this.kafkaVersion != null && this.kafkaVersion.equals("2.8.2")) || this.imageNameProvider.get().contains("2.8.2"))) {
```
tombentley, line 119: Perhaps we should use `startsWith("2.")`?

see-quick: Yeap, I thought about it but did not want to do it in this PR. But I will do it.

scholzj (approved):

tombentley (approved):

PR #58, opened 2023-06-27
Make maven sonatype plugin work also with Java 17, and publish for 17 only

see-quick:
The older sonatype maven plugin (`1.6.3`) is failing when pushing artefacts using Java 17. This update should solve such a problem with the latest version `1.6.13`.

Basically a mirror of the https://github.com/strimzi/strimzi-kafka-operator/pull/7946. 

Plus I have added for flaky test cases (that it well try at least 3 times if they end up failing) 
Plus, publishing artefacts will done only in scope of Java 17 build

### DIFF
--- a/.azure/build-pipeline.yaml
+++ b/.azure/build-pipeline.yaml
@@ -42,7 +42,7 @@ jobs:
         parameters:
           JDK_VERSION: $(jdk_version)
 
-      - bash: mvn clean verify
+      - bash: mvn -Dfailsafe.rerunFailingTestsCount=3 clean verify
         env:
           # Test container optimization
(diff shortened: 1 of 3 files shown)


scholzj (commented):
I don't think this solves the issue with the deployment happning in all Java versions. You should solve it together with this I think to avoid it being published twice next time.

see-quick:
> I don't think this solves the issue with the deployment happening in all Java versions. You should solve it together with this I think to avoid it being published twice next time.

Done. Publishing artefacts will be done only in the scope of the Java 17 build pipeline.

Frawless (approved):

scholzj (approved):

PR #62, opened 2023-11-11
bump test containers version

see-quick:
This PR adds a new version of test containers to pom.xml

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,7 +89,7 @@
         <maven.compiler.source>11</maven.compiler.source>
 
         <!-- DEPENDENCY CODE VERSIONS       -->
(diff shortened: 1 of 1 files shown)


scholzj (commented):
The tests don't seem to pass?

see-quick:
> The tests don't seem to pass?

Yeah, there are a few problems with deprecations, which should be replaced... will take a look on Monday.

fvaleri (approved):
LGTM. I would just replace the test ports (i.e. 9092) with a random local port to avoid clashing with common Kafka ports (a dedicated PR is fine for that).

scholzj (approved):
LGTM assuming it works.

Issue #72, opened 2024-07-04
Generated cluster id for kraft mode is sometimes invalid


fedinskiy:
When running a container in kraft mode, I sometimes (approximately once in 100 runs) see the container fail with status 1 and the following error in the logs:

```
2024-07-04T15:24:09.5338779Z usage: kafka-storage format [-h] --config CONFIG --cluster-id CLUSTER_ID
2024-07-04T15:24:09.5339351Z                      [--add-scram ADD_SCRAM] [--ignore-formatted]
2024-07-04T15:24:09.5339892Z                      [--release-version RELEASE_VERSION]
2024-07-04T15:24:09.5340442Z kafka-storage: error: argument --cluster-id/-t: expected one argument
2024-07-04T15:24:09.5341500Z [2024-07-04 15:19:11,430] INFO Registered kafka:type=kafka.Log4jController MBean (kafka.utils.Log4jControllerRegistration$)
2024-07-04T15:24:09.5343120Z [2024-07-04 15:19:11,654] INFO Setting -D jdk.tls.rejectClientInitiatedRenegotiation=true to disable client-initiated TLS renegotiation (org.apache.zookeeper.common.X509Util)
2024-07-04T15:24:09.5344263Z [2024-07-04 15:19:11,722] ERROR Exiting Kafka due to fatal exception (kafka.Kafka$)
2024-07-04T15:24:09.5344897Z java.lang.RuntimeException: No readable meta.properties files found.
2024-07-04T15:24:09.5345714Z 	at org.apache.kafka.metadata.properties.MetaPropertiesEnsemble.verify(MetaPropertiesEnsemble.java:493)
2024-07-04T15:24:09.5346567Z 	at kafka.server.KafkaRaftServer$.initializeLogDirs(KafkaRaftServer.scala:152)
2024-07-04T15:24:09.5347197Z 	at kafka.server.KafkaRaftServer.<init>(KafkaRaftServer.scala:60)
2024-07-04T15:24:09.5347787Z 	at kafka.Kafka$.buildServer(Kafka.scala:82)
2024-07-04T15:24:09.5348162Z 	at kafka.Kafka$.main(Kafka.scala:90)
2024-07-04T15:24:09.5348486Z 	at kafka.Kafka.main(Kafka.scala)
2024-07-04T15:24:09.5348701Z 
```

[...]

PR #73, opened 2024-07-04
Fix KRaft storage formating with UUID starting with dash

see-quick:
This PR solves [1] issue. TLDR; when we generate UUID starting with `-.....` the command treats it as a new parameter and therefore fails. 
  
[1] - https://github.com/strimzi/test-container/issues/72

### DIFF
--- a/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
+++ b/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
@@ -278,7 +278,7 @@ public class StrimziKafkaContainer extends GenericContainer<StrimziKafkaContaine
             command += "bin/kafka-server-start.sh config/server.properties" + kafkaConfigurationOverride;
         } else {
             clusterId = this.randomUuid();
-            command += "bin/kafka-storage.sh format -t= \"" + clusterId + "\" -c config/kraft/server.properties \n";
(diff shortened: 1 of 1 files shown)


src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -278,7 +278,7 @@ protected void containerIsStarting(final InspectContainerResponse containerInfo,
             command += "bin/kafka-server-start.sh config/server.properties" + kafkaConfigurationOverride;
         } else {
             clusterId = this.randomUuid();
-            command += "bin/kafka-storage.sh format -t \"" + clusterId + "\" -c config/kraft/server.properties \n";
+            command += "bin/kafka-storage.sh format -t= \"" + clusterId + "\" -c config/kraft/server.properties \n";
```
scholzj, line 281: Should it be without the space?

```suggestion
            command += "bin/kafka-storage.sh format -t=\"" + clusterId + "\" -c config/kraft/server.properties \n";
```

see-quick: yes correct :) thanks

scholzj (approved):

im-konge (approved):

Issue #63, opened 2023-12-28
Bundle Stirmzi Oauth in the Test Container containers to allow testing of OAuth


scholzj:
In some situation, it might be useful to have the Strimzi Oauth library packaged into the Test Container images as well to allow testing against it in tests.

see-quick:
I am thinking about it. So this is current set of images:
```java
REPOSITORY                   TAG                          IMAGE ID       CREATED              SIZE
strimzi/test-container       latest-kafka-3.6.1-ppc64le   05365a653b6b   2 seconds ago        491MB
strimzi/test-container       latest-kafka-3.6.1-arm64     af219b19bfa9   3 seconds ago        482MB
strimzi/test-container       latest-kafka-3.6.1-s390x     edc2318f6d2c   5 seconds ago        443MB
strimzi/test-container       latest-kafka-3.6.1-amd64     83ec1831ae31   7 seconds ago        471MB
strimzi/test-container       latest-kafka-3.5.1-ppc64le   2f53604bf871   10 seconds ago       484MB
strimzi/test-container       latest-kafka-3.5.1-arm64     aa6c24270fc4   12 seconds ago       475MB
strimzi/test-container       latest-kafka-3.5.1-s390x     6219d2bd8a6b   13 seconds ago       435MB
strimzi/test-container       latest-kafka-3.5.1-amd64     0f64674b283f   15 seconds ago       464MB
strimzi/test-container       latest-kafka-3.4.1-ppc64le   bb21a9df0291   18 seconds ago       484MB
strimzi/test-container       latest-kafka-3.4.1-arm64     0c637e04c345   20 seconds ago       474MB
strimzi/test-container       latest-kafka-3.4.1-s390x     ac24b93439ea   21 seconds ago       435MB
strimzi/test-container       latest-kafka-3.4.1-amd64     d8ba6ff1c276   23 seconds ago       463MB
strimzi/test-container       latest-kafka-3.3.2-ppc64le   4a6326322af5   26 seconds ago       484MB
strimzi/test-container       latest-kafka-3.3.2-arm64     b2765d349e3e   28 seconds ago       474MB
strimzi/test-container       latest-kafka-3.3.2-s390x     75d7966245ab   30 seconds ago       435MB
strimzi/test-container       latest-kafka-3.3.2-amd64     4e71bed5898e   31 seconds ago       463MB
strimzi/test-container       latest-kafka-3.2.3-ppc64le   333342b8b345   34 seconds ago       481MB
[...]

PR #76, opened 2024-08-15
Transfer only contents of Kafka and ZooKeeper properties files

MikeEdgar:
I've been playing around with this library to investigate listener configurations (likely helpful to #63), but I am unable to run the tests in an environment with podman. It turns out that copying a `MountableFile` to the containers preserves the user and group Ids from the host machine via an intermediate tar file. This fails with podman due to the chown command not being allowed.

The change proposed here is to read the contents of the `MountableFile` prior to copying into the container. This drops the user and group information, which should not be relevant. 

An additional change is to delay copying the Kafka properties until the container is starting, which allows for `withKraft` and `withServerProperties` to be called in any order.


see-quick (approved):
Thanks.

Issue #84, opened 2024-10-14
Improve `StrimziKafkaCluster` class to use of Builder pattern


see-quick:
Currently, the `StrimziKafkaCluster` class has multiple constructors, and adding more parameters requires creating additional constructors, which can negatively impact the user experience. We should implement the Builder pattern within this class to make it more flexible and better prepared for changes, such as when we need to expose more parameters.

PR #86, opened 2024-10-14
`StrimziKafkaCluster` builder

see-quick:
This PR adds flexibility to `StrimziKafkaCluster` by adapting the Builder pattern. With such change we will easily extend such class adding more parameters. Solves #84.

### DIFF
--- a/src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
+++ b/src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
@@ -203,7 +203,6 @@ public class StrimziKafkaCluster implements KafkaContainer {
         private Map<String, String> additionalKafkaConfiguration = new HashMap<>();
         private ToxiproxyContainer proxyContainer;
         private boolean enableSharedNetwork;
-        private String kafkaVersion;
 
         /**
          * Sets the number of Kafka brokers in the cluster.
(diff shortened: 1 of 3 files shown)


scholzj (commented):
I think the builder pattern is useful. But by removing the original constructors, you are breaking the compatibility. Is that wise / intended? Should we keep them in some way?

see-quick:
> I think the builder pattern is useful. But by removing the original constructors, you are breaking the compatibility. Is that wise / intended? Should we keep them in some way?

Hmmm, I was thinking about it but adding more and more constructors would end up with a lot of constructors. Maybe a less aggressive way would be having both (but `@Deprecate` constructors)?

scholzj:
I think deprecating them would be fine. The question is if you can combine both the deprecated constructors and the builders to allow users to transition smoothly.

see-quick:
> I think deprecating them would be fine. The question is if you can combine both the deprecated constructors and the builders to allow users to transition smoothly.

Yeah, it would be okay I think. It would be a matter of extracting the same code into methods so we don't end up with repetitive code. I will update this PR ten :)

src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
```diff
@@ -147,10 +126,153 @@ public StrimziKafkaCluster(final int brokersNum, final ToxiproxyContainer proxyC
...
+            .withNetwork(this.network);
+
+        if (this.proxyContainer != null) {
+            this.proxyContainer.setNetwork(this.network);
+        }
+
+        startKafkaCluster(this.additionalKafkaConfiguration);
+    }
+
+    private void startKafkaCluster(final Map<String, String> additionalKafkaConfiguration) {
```
scholzj, line 155: Does this actually start it? doesn't the `start()` method do it? Should it be maybe called `prepareKafkaCluster` or something?

see-quick: Oh yea 🤦 ...`prepareKafkaCluster` sounds better. Thanks

scholzj (approved):
Thanks.

PR #89, opened 2024-10-16
`StrimziKafkaCluster` KRaft mode

see-quick:
This PR adds support for KRaft within `StrimziKafkaCluster`. Currently, we are able to run in KRaft mode with the single-node instance.

### DIFF
--- a/.checkstyle/checkstyle.xml
+++ b/.checkstyle/checkstyle.xml
@@ -98,7 +98,7 @@
         </module>
         <module name="CyclomaticComplexity">
             <!-- default is 10-->
-            <property name="max" value="11"/>
+            <property name="max" value="13"/>
         </module>
         <module name="JavaNCSS"/>
         <module name="NPathComplexity">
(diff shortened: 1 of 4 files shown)


.checkstyle/checkstyle.xml
```diff
@@ -98,7 +98,7 @@
         </module>
         <module name="CyclomaticComplexity">
             <!-- default is 10-->
-            <property name="max" value="11"/>
```
scholzj, line 101: If you need to override this for some method, you should use annotation and not change the rules for everything.

see-quick: Thanks. I simplify code a little bit and this is not needed anymore :)

scholzj (approved):
LGTM assuming it works 😉. At some poinbt we should also update the README. But that can be done in separate PR. Thanks.

jankalinic (approved):

see-quick:
> LGTM assuming it works 😉. At some poinbt we should also update the README. But that can be done in separate PR. Thanks.

Yeah, I will update README in separate PR.

PR #91, opened 2024-10-16
Add KRaft readiness check in StrimziKafkaCluster

see-quick:
This PR adds a KRaft readiness check within `StrimziKafkaCluster`. Currently `waitForRunning()` inside each `StrimziKafkaContainer` is not enough because it does not wait on Quorum form but only for `Transitioning from RECOVERY to RUNNING` which is only correct when we use single-node. In case of multi-node we need to wait until quorum is formed and ready.

### DIFF
--- a/src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
+++ b/src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
@@ -20,6 +20,8 @@ import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
@@ -382,6 +384,7 @@ public class StrimziKafkaCluster implements KafkaContainer {
         additionalKafkaConfiguration.put("controller.quorum.voters", quorumVoters);
     }
 
+    @SuppressWarnings({"CyclomaticComplexity"})
     @Override
     public void start() {
         Stream<KafkaContainer> startables = this.brokers.stream();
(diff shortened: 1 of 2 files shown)


im-konge (approved):

scholzj (approved):
LGTM if it works

src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
```diff
@@ -412,6 +415,50 @@ public void start() {
                         return false;
                     }
                 });
+        } else if (this.isKraftKafkaCluster()) {
+            // Readiness check for KRaft mode
+            Utils.waitFor("Kafka brokers to form a quorum", Duration.ofSeconds(5).toMillis(), Duration.ofMinutes(1).toMillis(),
```
Frawless, line 420: isn't 5 sec too much for pooling?

scholzj: Good catch ... we probably want it to be fairly aggressive.

see-quick: Well, I can try 2 seconds. WDYT @Frawless? The tests are no longer flaky with this change
```java
Tests run: 177, Failures: 0, Errors: 0
```

see-quick: This is with 5s polling so I will try 1s and see how time execution will change.
```
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  21:22 min
[INFO] Finished at: 2024-10-16T12:28:53Z
```

scholzj: I would go to 1 second I guess.

scholzj: Or make it configurable 😉

see-quick:
```
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  20:05 min
[INFO] Finished at: 2024-10-16T13:13:17Z
```
For 1s :) That's a solid improvement.

Frawless (approved):

PR #92, opened 2024-10-16
Optimize a bit memory usage of StrimziKafkaContainer

see-quick:
This PR tries to solve problems with memory consumption. 
```
##[warning]Free memory is lower than 5%; Currently used: 95.10%
##[warning]Free memory is lower than 5%; Currently used: 95.10%
```


scholzj (commented):
Do we really want to have change the default here? I think they should be chngeable by the users. but changing it like this could break a lot of things without a clear reason.

see-quick:
> Do we really want to have change the default here? I think they should be chngeable by the users. but changing it like this could break a lot of things without a clear reason.

Yeah, that's reasonable.

PR #99, opened 2024-11-04
Fix memory problems inside test suite

see-quick:
In the Add proxy container configuration (#44), we forgot to add stop(). That causes containers to run during all other test cases. Only recently (with more Kafka versions) it showed that we are out of memory in Azure CI:
```java
##[warning]Free memory is lower than 5%; Currently used: 95.10%
##[warning]Free memory is lower than 5%; Currently used: 95.10%
```

This should fix the problem above.

Note: 

I think we should double-check before each test that none containers are running to avoid this in the future.

### DIFF
--- a/src/test/java/io/strimzi/test/container/StrimziKafkaContainerIT.java
+++ b/src/test/java/io/strimzi/test/container/StrimziKafkaContainerIT.java
@@ -17,6 +17,7 @@ import org.apache.kafka.clients.producer.ProducerConfig;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.common.serialization.StringDeserializer;
 import org.apache.kafka.common.serialization.StringSerializer;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.MethodSource;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


src/test/java/io/strimzi/test/container/StrimziKafkaContainerIT.java
```diff
@@ -312,6 +312,7 @@ void testGetProxyWithNoContainer(final String imageName) {
                 .waitForRunning();
         systemUnderTest.start();
         assertThrows(IllegalStateException.class, () -> systemUnderTest.getProxy());
+        systemUnderTest.stop();
```
Frawless, line 315: Wouldn't be better to put it into `afterEach` ?

see-quick: > Wouldn't be better to put it into `afterEach` ?

Some tests doesn't to start but I think that we can solve it like `if sut != null -> sut.stop()`.

Frawless: Ye that was exactly the proposed solution I was thinking about. It is just an idea, if you think it will be better to just add it into one test I am fine with it.

see-quick: Like second option would be to re-write all of those tests to `try-with-resources` as this class implements AutoClosable.

see-quick: Update both for ZK and KRaft suite with afterEach.

Frawless (approved):

src/test/java/io/strimzi/test/container/StrimziKafkaContainerIT.java
im-konge: so should this be removed or it should be kept?

PR #104, opened 2024-11-05
Mutation testing 

see-quick:
This PR adds to the test container mutation testing. Specifically [1]. During adoption, I saw that there are a lot of paths in the `Strimzi*` classes, which are not tested. So I have also added a few tests to solve it. 

You can read about the benefits of mutation testing here [2].

Moreover, I have updated Azure-CI with mutation testing within the pipeline. Also if someone writes a new code then the build may fail because I have set the following thresholds. So there will be a need to add UTs to cover such features and be at least on such threshold as we have currently.
```
<mutationThreshold>80</mutationThreshold>
<coverageThreshold>71</coverageThreshold>
```

[1] - https://pitest.org/
[2] - https://www.arcmutate.com/benefits/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -431,7 +431,8 @@
                         <param>io.strimzi.test.container.Strimzi*</param>
                     </targetClasses>
                     <targetTests>
-                        <param>io.strimzi.test.container.Strimzi*</param>
+                        <!-- include just UTs within mutation and prevent image pulling -->
+                        <param>io.strimzi.test.container.Strimzi*Test</param>
                     </targetTests>
                     <avoidCallsTo>
                         <avoidCallsTo>java.util.logging</avoidCallsTo>
@@ -439,7 +440,7 @@
                         <avoidCallsTo>org.slf4j</avoidCallsTo>
                         <avoidCallsTo>org.apache.commons.logging</avoidCallsTo>
                     </avoidCallsTo>
-                    <mutationThreshold>82</mutationThreshold>
+                    <mutationThreshold>80</mutationThreshold>
                     <coverageThreshold>73</coverageThreshold>
                     <verbose>true</verbose>
                 </configuration>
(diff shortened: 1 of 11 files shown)


im-konge (approved):

pom.xml
```diff
@@ -398,6 +414,37 @@
...
+                    <dependency>
+                        <groupId>org.pitest</groupId>
+                        <artifactId>pitest-junit5-plugin</artifactId>
+                        <version>${pit-junit-plugin.version}</version>
+                    </dependency>
+                </dependencies>
+                <configuration>
+                    <targetClasses>
+                        <!-- we not want to include also WaitException, Utils classes etc. -->
+                        <param>io.strimzi.test.container.Strimzi*</param>
```
Frawless, line 431: What about `KafkaContianer` and `KafkaServiceVersion`? They are not needed as well?

Frawless (approved):
I am okay with it, but using `@DoNotMutate` to quite a lot of methods within the class looks a little bit strange.

see-quick:
> I am okay with it, but using `@DoNotMutate` to quite a lot of methods within the class looks a little bit strange.

Yeah, but those methods are mainly used in Integration tests or cause a timeout of Mutation minions (e.g., randomUuid). In general, you don't want to include those because I would need to add these tests to UTs and that's not a good approach IMO (f.e., image pulling). Because now they run around 50s and with those tests, it would need like much more...

pom.xml
see-quick: Hmm KafkaContianer is just interface but fair enough if we add some default methods there then it would make sense. I will update it. And for  `KafkaServiceVersion` I didn't want to include and focus on main classes but if you think it would make sense I can add it. :)

Frawless: Maybe more explanation in comment would be good, I am fine with any change you prefer as SME.

PR #110, opened 2024-11-11
Add Surefire plugin

see-quick:
This PR adds a surefire plugin to run UTs within the test container.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -110,7 +110,7 @@
         <!-- DEPENDENCY TEST VERSIONS      -->
         <jupiter.version>5.11.0</jupiter.version>
         <hamcrest.version>2.2</hamcrest.version>
-        <maven-failsafe-plugin.version>3.0.0-M7</maven-failsafe-plugin.version>
+        <maven-failsafe-plugin.version>3.1.2</maven-failsafe-plugin.version>
         <maven-surefire-plugin.version>3.1.2</maven-surefire-plugin.version>
         <!-- for Java 11 compatibility (newer versions does not support Java 11)  -->
         <testcontainers-keycloak.version>2.6.0</testcontainers-keycloak.version>
--- a/pom.xml
+++ b/pom.xml
@@ -110,8 +110,7 @@
         <!-- DEPENDENCY TEST VERSIONS      -->
         <jupiter.version>5.11.0</jupiter.version>
         <hamcrest.version>2.2</hamcrest.version>
-        <maven-failsafe-plugin.version>3.1.2</maven-failsafe-plugin.version>
-        <maven-surefire-plugin.version>3.1.2</maven-surefire-plugin.version>
+        <maven-surefire-failsafe-plugin.version>3.1.2</maven-surefire-failsafe-plugin.version>
         <!-- for Java 11 compatibility (newer versions does not support Java 11)  -->
         <testcontainers-keycloak.version>2.6.0</testcontainers-keycloak.version>
         <fasterxml.jackson-annotations.version>2.16.2</fasterxml.jackson-annotations.version>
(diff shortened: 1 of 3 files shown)


im-konge (approved):

src/test/java/io/strimzi/test/container/AbstractIT.java
```diff
@@ -41,7 +41,9 @@ protected static Stream<Arguments> retrieveKafkaVersionsFile() throws IOExceptio
     }
 
     protected void assumeDocker() {
-        Assumptions.assumeTrue(System.getenv("DOCKER_CMD") == null || "docker".equals(System.getenv("DOCKER_CMD")));
+        Assumptions.assumeTrue(
+            System.getenv("DOCKER_CMD") == null ||
+                ("docker".equals(System.getenv("DOCKER_CMD")) || "podman".equals(System.getenv("DOCKER_CMD"))));
```
Frawless, line 46: doesn't it mean - always true?

see-quick: Maybe having just `System.getenv("DOCKER_CMD") == null` should be okay? I don't want to be more strict (i.e., 
```java
Assumptions.assumeTrue(
    System.getenv("DOCKER_CMD") != null &&
        ("docker".equals(System.getenv("DOCKER_CMD")) || "podman".equals(System.getenv("DOCKER_CMD"))));
```

scholzj, line 46: What exactly is the purpose of this? Why do we check some environment variable like this?

Frawless: Hard to say, what is the reason for this? It is for `docker` only tests? Or `docker` has to be installed ?

see-quick: > It is for docker only tests? Or docker has to be installed ?

Both I think.

see-quick: Well as I said I am not sure what was the first intention with this. I can remove this easily... :). Because I also don't see any strong benefit of such check...

Frawless: +1 for removing it

Frawless: Let's move the discussion bellow.

scholzj: Yeah ... if we do not know why we have it and it doesn't seem to make sense, let's remove it I guess.

see-quick: Removed.

scholzj (approved):
So, in all other projects we run the unit and integration tests with one step. Not sure if there is any reason to do it separately. If there is some reason for it and you want to keep it separate, its fine with me.

see-quick:
> So, in all other projects we run the unit and integration tests with one step. Not sure if there is any reason to do it separately. If there is some reason for it and you want to keep it separate, its fine with me.

Hmmm, I wanted to do it simultaneously but now that I have mutation testing  included I want to first check:
1. Unit tests (if that works move to the next point)
2. Mutation tests where I check coverage (i.e., mutation and line coverage) from Unit tests (if someone didn't cover a feature with Unit tests mutation/line coverage would fail otherwise one can provide to the next point)
 3. Run integration tests

scholzj:
Ok. As I said, if it has some reason, it is fine with with me.

pom.xml
```diff
@@ -111,6 +111,7 @@
         <jupiter.version>5.11.0</jupiter.version>
         <hamcrest.version>2.2</hamcrest.version>
         <maven-failsafe-plugin.version>3.0.0-M7</maven-failsafe-plugin.version>
+        <maven-surefire-plugin.version>3.1.2</maven-surefire-plugin.version>
```
Frawless, line 114: Why are you using different versions for failsafe and surefire?

see-quick: nothing intentional. I have just use some version and didn't want to change more than it is needed. But as you pointed it out I have updated also failsafe plugin version to match surefire. Thanks

Frawless (approved):

PR #111, opened 2024-11-12
Add Controller listener to the advertised.listeners

see-quick:
This PR adds the `CONTROLLER` to `advertised.listeners`.

### DIFF
--- a/src/main/java/io/strimzi/test/container/KafkaVersionService.java
+++ b/src/main/java/io/strimzi/test/container/KafkaVersionService.java
@@ -139,6 +139,29 @@ class KafkaVersionService {
         }
 
         /**
+         * Extracts the Kafka version from a Docker image name.
+         *
+         * <p>Expects the image name to contain "-kafka-" followed by the version number at the end.
+         * For example:
+         * <ul>
+         *   <li>"quay.io/strimzi-test-container/test-container:0.107.0-rc1-kafka-3.7.1" returns "3.7.1"</li>
+         *   <li>"quay.io/strimzi-test-container/test-container:0.105.0-kafka-3.6.0" returns "3.6.0"</li>
+         * </ul>
+         *
+         * @param imageName the Docker image name (e.g., "quay.io/...:0.107.0-rc1-kafka-3.7.1")
+         * @return the extracted Kafka version (e.g., "3.7.1")
+         * @throws IllegalArgumentException if the version cannot be extracted
+         */
+        public static String extractVersionFromImageName(final String imageName) {
+            final Matcher matcher = KAFKA_VERSION_PATTERN.matcher(imageName);
+            if (matcher.find()) {
+                return matcher.group(1); // Returns the Kafka version string like "3.9.0"
+            } else {
+                throw new IllegalArgumentException("Cannot extract Kafka version from image name: " + imageName);
+            }
+        }
+
+        /**
          * Get the Kafka version in the following format (i.e., "3.0.0")
          *
          * @return kafka version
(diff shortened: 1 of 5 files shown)


scholzj (commented):
I think you need to be careful with this. You want to listen on 0.0.0.0. But you do not want to advertise it. That does not seem to be what you are doing here.

see-quick:
> I think you need to be careful with this. You want to listen on 0.0.0.0. But you do not want to advertise it. That does not seem to be what you are doing here.

I see, currently, we have in single-node this configuration
```
advertised.listeners=PLAINTEXT\://localhost\:33219,BROKER1\://10.90.209.2\:9091
broker.id=1
controller.listener.names=CONTROLLER
controller.quorum.voters=1@broker-1\:9094
inter.broker.listener.name=BROKER1
listener.security.protocol.map=PLAINTEXT\:PLAINTEXT,CONTROLLER\:PLAINTEXT,BROKER1\:PLAINTEXT
listeners=PLAINTEXT\://0.0.0.0\:9092,BROKER1\://0.0.0.0\:9091,CONTROLLER\://broker-1\:9094
...
```
for the controller listener, I had to change it to DNS [1] i.e., `CONTROLLER\://broker-1\:9094`. But I assume instead of `BROKER1\://10.90.209.2\:9091` I can use `BROKER1\://broker-1\:9091`.  That way for instance producer or consumer would easily connect to the internal docker network by just specifying `BROKER1\://broker-1\:9091` as bootstrap servers. External clients have to use `PLAINTEXT\://localhost\:33219`.

[1] - https://github.com/apache/kafka/commit/9be27e715a209a892941bf35e66859d9c39c28c4

scholzj:
No. You need to add the controller to the advertised hosts.

src/main/java/io/strimzi/test/container/KafkaVersionService.java
```diff
@@ -135,6 +138,29 @@ public static int compareVersions(String version1, String version2) {
...
+         * </ul>
+         *
+         * @param imageName the Docker image name (e.g., "quay.io/...:0.107.0-rc1-kafka-3.7.1")
+         * @return the extracted Kafka version (e.g., "3.7.1")
+         * @throws IllegalArgumentException if the version cannot be extracted
+         */
+        public static String extractVersionFromImageName(final String imageName) {
+            final Matcher matcher = KAFKA_VERSION_PATTERN.matcher(imageName);
+            if (matcher.find()) {
+                return matcher.group(1); // Returns the Kafka version string like "3.9.0"
```
scholzj, line 158: This is pretty unreliable. I think we need to have better way to configure KAfka version than through the container image. If you do not want to solve it in this PR it might be fine, but we should have issues to track this.

This is pretty likel to fail when using custom images when in various ofline / disconnected environments, when you want to fix to use some specific image through digest etc.

see-quick: Yeah, I was thinking about the same with custom images. I can create an issue when I solve that also for custom images or at least log a warning when using custom images.

scholzj: I think it should work like this:
* The builders should have a `withVersion` field/method
* If not set, the default version (the latest Kafka version) should be used
* The image should be by the users always set to correspond to the version

I think that is the way it should work. (As I said, I'm happy to leave it for later and not block this PR.)

scholzj (approved):

src/main/java/io/strimzi/test/container/KafkaVersionService.java
see-quick: Yeah, I think I have come up with a more clear solution. Let me know what you think

PR #114, opened 2024-11-14
Various Java cleanups

mimaison:
- Removed usage of deprecated APIs
- Minor refactorings and code simplifications

### DIFF
--- a/src/main/java/io/strimzi/test/container/KafkaVersionService.java
+++ b/src/main/java/io/strimzi/test/container/KafkaVersionService.java
@@ -85,7 +85,7 @@ class KafkaVersionService {
                 }
                 throw new UnknownKafkaVersionException("Doesn't know the specified Kafka version: " + kafkaVersion + ". " +
                     "The supported Kafka versions are: " +
-                    KafkaVersionService.getInstance().logicalKafkaVersionEntities.stream().map(KafkaVersion::getVersion).collect(Collectors.toList()).toString());
+                    KafkaVersionService.getInstance().logicalKafkaVersionEntities.stream().map(KafkaVersion::getVersion).collect(Collectors.toList()));
             }
         }
         return imageName;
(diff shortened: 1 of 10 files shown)


src/main/java/io/strimzi/test/container/KafkaVersionService.java
```diff
@@ -197,7 +197,7 @@ public String toString() {
      */
     public KafkaVersion latestRelease() {
         // at least one release in the json schema is needed
-        if (this.logicalKafkaVersionEntities == null || this.logicalKafkaVersionEntities.size() < 1) {
+        if (this.logicalKafkaVersionEntities.isEmpty()) {
```
mimaison, line 200: The field is `final` and initialized when declared so it should never be `null`.

src/test/java/io/strimzi/test/container/StrimziKafkaContainerIT.java
```diff
@@ -72,7 +72,7 @@ void testStartContainerWithEmptyConfiguration(final String imageName) {
             systemUnderTest.start();
 
             assertThat(systemUnderTest.getBootstrapServers(), is("PLAINTEXT://"
-                    + systemUnderTest.getContainerIpAddress() + ":" + systemUnderTest.getMappedPort(9092)));
+                    + systemUnderTest.getHost() + ":" + systemUnderTest.getMappedPort(9092)));
```
mimaison, line 75: `getContainerIpAddress()` is deprecated, so switching to `getHost()` which is the preferred method now.

src/test/java/io/strimzi/test/container/StrimziKafkaContainerMockTest.java
```diff
@@ -18,8 +18,8 @@
 import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.junit.Assert.assertTrue;
```
mimaison, line 21: This is the junit 4 dependency brought via testcontainers. We should instead use `org.junit.jupiter.api.Assertions` which is junit 5 (the declared dependency in pom.xml).

see-quick (approved):
Thanks, please update the description of this PR.

mimaison:
Thanks, I updated the description.

Issue #100, opened 2024-11-04
Add precondition to Integration tests that none containers are running 


see-quick:
I think we should add a precondition to each test case (i.e., on the top of the `AbstractIT` class `@BeforeEach`) that none containers are running. This would prevent issues like this https://github.com/strimzi/test-container/pull/99.

PR #118, opened 2024-11-26
Removes finally blocks and add @AfterEach to each ITs to stop running container

see-quick:
This PR fixes and removed the `finally` block and instead added the @AfterEach annotation to IT suites that were missing it.

Solves #100


scholzj:
I don't follow this. Why should there be no running containers?

see-quick:
> I don't follow this. Why should there be no running containers?

Each test case has its resources. Currently, we do not use shared resources across more tests. As I have tested now, there is no need to use that precondition and only `@AfterEach` is enough.

im-konge (approved):

jankalinic (approved):

PR #122, opened 2024-12-02
Fix Intermittent Kafka Quorum Formation Issue

see-quick:
This PR addresses an intermittent issue where Kafka brokers fail to form a quorum due to mismatched voter keys in the Raft consensus algorithm. The problem occurs approximately 1 out of every 5 test runs (which I tried)...

Brokers occasionally fail to form a quorum, with logs showing repeated messages like:
```java
Leader sent a voter key... that doesn't match the local key...; returning INVALID_VOTER_KEY
```

The advertised listeners for the CONTROLLER listener are set to localhost (using `getHost()` method), while controller.quorum. voters specify hostnames like broker-0, broker-1, etc. This mismatch leads to brokers advertising their controller endpoints incorrectly, causing other brokers to attempt connections to localhost instead of the correct broker addresses.

Using localhost for the CONTROLLER listener causes brokers to incorrectly attempt to connect to themselves rather than the intended broker, due to localhost resolving to 127.0.0.1.

I have verified such a change in a test case which failed once in 5 runs and now it's working (20 runs without any issue):

image

### DIFF
--- a/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
+++ b/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
@@ -274,6 +274,7 @@ public class StrimziKafkaContainer extends GenericContainer<StrimziKafkaContaine
                 LOGGER.info("New `cluster.id` has been generated: {}", this.clusterId);
             }
 
+            command += "cat /opt/kafka/config/kraft/server.properties\n";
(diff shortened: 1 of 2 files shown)


showuon (approved):
Nice find @see-quick ! The fix makes sense to me.

henryZrncik (approved):

im-konge (approved):

PR #123, opened 2024-12-06
Add option to enable logging into stdout using env variable

see-quick:
This PR adds an option to enable logging into stdout using the env variable `STRIMZI_TEST_CONTAINER_LOGGING_ENABLED`. 

An example of the log can be found [1].

[1] - https://gist.githubusercontent.com/see-quick/8d88a97b533f8910ec2af135071b6ab7/raw/57e6cc73e8a875de75e0655677fac30bba5b67f8/gistfile1.txt

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,7 +235,7 @@ StrimziKafkaCluster kafkaCluster = new StrimziKafkaCluster.StrimziKafkaClusterBu
 kafkaCluster.start();
 ```
 
-#### x) Logging Kafka Container/Cluster Output to SLF4J
+#### xii) Logging Kafka Container/Cluster Output to SLF4J
 
 If you want to enable logging of the Kafka container’s output to SLF4J, 
 you can set the environment variable STRIMZI_TEST_CONTAINER_LOGGING_ENABLED to true. 
--- a/README.md
+++ b/README.md
@@ -235,6 +235,13 @@ StrimziKafkaCluster kafkaCluster = new StrimziKafkaCluster.StrimziKafkaClusterBu
 kafkaCluster.start();
 ```
 
+#### x) Logging Kafka Container/Cluster Output to SLF4J
+
+If you want to enable logging of the Kafka container’s output to SLF4J, 
+you can set the environment variable STRIMZI_TEST_CONTAINER_LOGGING_ENABLED to true. 
+By default, this feature is disabled.
+This can help in debugging or monitoring the Kafka broker’s activities during tests.
+
 ### Running Mutation Testing
 
 To run mutation testing and assess your code’s robustness against small changes, use the following Maven command:
--- a/README.md
+++ b/README.md
@@ -206,7 +206,7 @@ systemUnderTest.start();
 strimziKafkaContainer.getProxy().setConnectionCut(true);
 ```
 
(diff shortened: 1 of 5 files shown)


scholzj (commented):
Why do you call it `logToConsole` in one place and `BrokerContainerSlf4jLogging` in the other? Would the same name everywhere make sense? Also, having javadocs explaining what it actually does would help. `console` has many meanings for example.

Also, would some environment variable or system property be better for a switch like that? If you aim this for troubleshooting, then you likely do not want to modify the tests to troubleshooth but for example just set an environment variable maybe?

Frawless:
> Also, would some environment variable or system property be better for a switch like that? If you aim this for troubleshooting, then you likely do not want to modify the tests to troubleshooth but for example just set an environment variable maybe?

Wouldn't be better to log it to file instead of to console? Guess it can pollute the log quite quickly in case user will have more than few containers.

see-quick:
> > Also, would some environment variable or system property be better for a switch like that? If you aim this for troubleshooting, then you likely do not want to modify the tests to troubleshooth but for example just set an environment variable maybe?
> 
> Wouldn't be better to log it to file instead of to console? Guess it can pollute the log quite quickly in case user will have more than few containers.

So in that case I think it would be best to call `getLogs()` (if something bad happens) and append it to the logs files inside `/target/strimzi-test-container/<date>` as we have for instance LogCollector.  

> Also, would some environment variable or system property be better for a switch like that? If you aim this for troubleshooting, then you likely do not want to modify the tests to troubleshooth but for example just set an environment variable maybe?

Yeah, this is a good point. I think it would be better to use ENV or system property here. The main reason it to just enhance current troubleshooting in CI.

see-quick:
Any opinions on this @scholzj @Frawless? I can start with just:
1. Add ENV to enable console output for the container.

That would be the easiest way possible.

Frawless:
> Any opinions on this @scholzj @Frawless? I can start with just:
> 
> 1. Add ENV to enable console output for the container.
> 
> That would be the easiest way possible.

Yes, sounds reasonable to me. 

I am thinking about option to force logging without env for certain situations/tests and I don't have strong opinion. I can imagine situations that might be useful, but user can also achieve ti with ENV option that will of course turn on logging for everything.

Frawless (approved):

scholzj (commented):
Maybe you should update the PR description and add something to the README.md file?

see-quick:
> Maybe you should update the PR description and add something to the README.md file?

Yeah, thanks for it. I have updated it.

README.md
```diff
@@ -235,6 +235,13 @@ StrimziKafkaCluster kafkaCluster = new StrimziKafkaCluster.StrimziKafkaClusterBu
 kafkaCluster.start();
 ```
 
+#### x) Logging Kafka Container/Cluster Output to SLF4J
```
scholzj, line 238: The text is good. But it looks like you have `ix`, `xi` and then `x`?

see-quick: updated

scholzj: Yeah, but you are still missing `x` -> the previous one should be `x` and this one `xi`?

Issue #124, opened 2024-12-20
Code checks old brokerId when changin it in KRaft mode


fedinskiy:
When I am running a container in KRaft mode I am also setting up a custom brokerid. Since version 0.108.0 it also requires to set up a nodeId. When I set up both[1],  the method fails with `IllegalState `broker.id` and `node.id` must have same value!`, and when I don't, it fails with `java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because "this.nodeId" is null`.

It seem the problem is in method `withBrokerId` ([source](https://github.com/strimzi/test-container/blob/main/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java#L656C1-L663C6)), which checks broker id **before** this id is about to change and has no null pointer check on nodeid. This check was added in [this PR](https://github.com/strimzi/test-container/pull/89): despite kraft mode working fine in 0.107.0

[1]
```
  .withNodeId(1)
                .withKraft()
                .withBrokerId(1)
```

see-quick:
Thanks. That's a problem and I think we should let `withBrokerId(...)` be simply just builder option without any validations because then it could lead to these `order-dependent configuration`. In this case (for now) I think you can be fine with just:
```java
  .withBrokerId(1)
  ```
  because AFAIK I am setting it (at least in 0.109.0 version).
  ``java
  if (this.nodeId == null) {
            LOGGER.warn("Node ID is not set. Using broker ID {} as the default node ID.", this.brokerId);
            this.nodeId = this.brokerId;
        }
```
and such validation should be done in the `doStart()` method to avoid this order-depedent configuration.

fedinskiy:
@see-quick I am preparing a fix right now, is the check should be in `doStart` or in `containerIsStarting`? You set nodeId in the latter, so it will make sense to do the check there as well

see-quick:
> @see-quick I am preparing a fix right now, is the check should be in doStart or in containerIsStarting? You set nodeId in the latter, so it will make sense to do the check there as well

Yeah in `containerIsStarting` it makes more sense to have that validation.

PR #125, opened 2024-12-20
Check node and broker id equality in KRaft mode on start up and not in broker.id setter

fedinskiy:
Fixes https://github.com/strimzi/test-container/issues/124

### DIFF
--- a/src/test/java/io/strimzi/test/container/StrimziKafkaKraftContainerIT.java
+++ b/src/test/java/io/strimzi/test/container/StrimziKafkaKraftContainerIT.java
@@ -20,6 +20,7 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.MethodSource;
+import org.testcontainers.containers.ContainerLaunchException;
 import org.testcontainers.shaded.com.google.common.collect.ImmutableMap;
 
 import java.time.Duration;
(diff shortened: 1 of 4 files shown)


src/test/java/io/strimzi/test/container/StrimziKafkaKraftContainerIT.java
```diff
@@ -128,6 +130,24 @@ void testUnsupportedKRaftUsingImageName() {
         assertThrows(UnsupportedKraftKafkaVersionException.class, () -> systemUnderTest.start());
     }
 
+    @Test
+    void testUNsupportedKraftAndIdsMismatch() {
+        systemUnderTest = new StrimziKafkaContainer("quay.io/strimzi-test-container/test-container:latest-kafka-2.8.2")
+                .withBrokerId(1)
+                .withNodeId(1)
```
see-quick, line 137: and I think you have to have different IDs to achieve such an exception, right?

see-quick:
Also, there are a few checkstyle errors i.e.,
```
[ERROR] /home/vsts/work/1/s/src/test/java/io/strimzi/test/container/StrimziKafkaKraftContainerIT.java:139:35: Empty statement. [EmptyStatement]
[ERROR] /home/vsts/work/1/s/src/test/java/io/strimzi/test/container/StrimziKafkaKraftContainerIT.java:141: 'lambda arguments' has incorrect indentation level 16, expected level should be 12. [Indentation]
```
before you push always try to execute `mvn clean install -DskipTests`.

see-quick:
Also as you removed this one test case the mutation testing is failing because the average now of what is covered is less than it was before. 
```java
Line coverage of 71(326/459) is below threshold of 72
```
So please update it also in the `pom.xml`. Specifically here [1].

[1] - https://github.com/strimzi/test-container/blob/main/pom.xml#L458

see-quick (approved):
Thanks

scholzj (approved):

PR #130, opened 2025-03-18
Add Kafka 4.0.0 and removal of ZK support

see-quick:
This PR adds Kafka 4.0.0 and removes all related code to ZK support. Moreover, it follows the support matrix of the operator repository (i.e., 3.9.x, 4.0.x).


im-konge (approved):

im-konge:
Btw you should mention it definitely in the release notes that we are not supporting ZK anymore.

PR #134, opened 2025-03-27
Remove unused useKraft field and associated comments

MikeEdgar:
Minor change to remove the now unused `useKraft` field and comments that reference it.


see-quick (approved):
Thanks

PR #137, opened 2025-04-24
Expose GenericContainer from StrimziKafkaCluster via getNodes() method

see-quick:
This PR renames brokers to nodes to reflect KRaft’s mixed roles and exposes GenericContainer instances via getNodes(), enabling direct access to container-level operations like port exposure and file mounting. Moreover, it deprecates our getBrokers() method and directly tells the end user to use getNodes().

### DIFF
--- a/src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
+++ b/src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
@@ -8,6 +8,7 @@ import com.groupcdg.pitest.annotations.DoNotMutate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.testcontainers.containers.Container;
+import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.Network;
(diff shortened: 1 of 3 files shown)


im-konge (approved):

src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
```diff
@@ -228,10 +229,24 @@ public StrimziKafkaCluster build() {
...
      * @return collection of Strimzi kafka containers
      */
+    @Deprecated
     public Collection<KafkaContainer> getBrokers() {
-        return this.brokers;
+        return this.nodes;
+    }
+
+    /**
+     * Returns the underlying GenericContainer instances for Kafka brokers.
```
mimaison, line 242: That also returns controller containers, right?

see-quick: Yeah, I would rephrase it a bit.

mimaison (approved):
LGTM

ppatierno (approved):

PR #149, opened 2025-06-20
chore(deps): commons-compress to follow test container version 

see-quick:
Downgrade the commons-compress to 1.24.0 to follow test-container versions and avoid bumping and overriding transitive dependencies.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -128,11 +128,6 @@
         <pit-junit-plugin.version>1.2.1</pit-junit-plugin.version>
         <pit-plugin.version>1.18.1</pit-plugin.version>
 
-        <!-- FIX VULNERABILITY VERSIONS  -->
-        <commons-compress.version>1.27.1</commons-compress.version>
-        <commons-codec.version>1.18.0</commons-codec.version>
-        <commons-lang3.version>3.16.0</commons-lang3.version>
-        <commons-io.version>2.16.1</commons-io.version>
         <junit4.version>4.13.2</junit4.version>
 
         <!-- properties to skip surefire tests during failsafe execution -->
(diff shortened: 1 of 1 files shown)


scholzj (changes requested):
Sorry, but I still don't understand the need for these changes.

scholzj:
I mean what is it you are actually trying to fix here?

scholzj:
But why are you updating it? Why don't you use what Test Container uses?

see-quick:
> But why are you updating it? Why don't you use what Test Container uses?

The Test Container uses a version that has CVE vulnerabilities.
[CVE-2024-26308](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2024-26308)
[CVE-2024-25710](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2024-25710)

scholzj:
> > But why are you updating it? Why don't you use what Test Container uses?
> 
> The Test Container uses a version that has CVE vulnerabilities: [CVE-2024-26308](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2024-26308) [CVE-2024-25710](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2024-25710)

Ok, so ... do these vulnerabilities really impact us? And given this is just a **Test** container which is supposed to be used in unit and integration tests, do we really care about them given the overriding does not seem to be simple? Not every CVE is worth fixing and it is not like we are (de)compressing some user data from the internet or something.

scholzj:
Haha, yeah. That sucks. I guess people change. :-D

see-quick:
@strimzi/maintainers What are your thoughts? (a) overrides (b) downgrade to 1.24.0?

im-konge (approved):

PR #153, opened 2025-09-11
Add support for KRaft separate roles

see-quick:
This PR adds support for KRaft separate roles (i.e., BROKER, CONTROLLER). So now you could deploy StrimziKafkaCluster with, for instance, 3 broker roles and 3 controller nodes instead of 6 mixed-nodes. The backwards compatibility is done by using MIXED mode by default. This may change in the future, but for now, I decided to have mixed mode by default as it was before.


src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -88,6 +89,7 @@ public class StrimziKafkaContainer extends GenericContainer<StrimziKafkaContaine
     private Function<StrimziKafkaContainer, String> bootstrapServersProvider = c -> String.format("PLAINTEXT://%s:%s", getHost(), this.kafkaExposedPort);
     private String clusterId;
     private MountableFile serverPropertiesFile;
+    private KafkaNodeRole nodeRole = KafkaNodeRole.COMBINED;   // backwards compatibility - we may by default use separate roles
```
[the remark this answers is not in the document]
see-quick: Currently, we are using mixed/combined mode by default => `backwards compatibility`. But in the future, that default could change.

src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
```diff
@@ -210,30 +291,42 @@ public StrimziKafkaClusterBuilder withKafkaVersion(String kafkaVersion) {
             return this;
         }
 
+        /**
+         * Configures the cluster to use separate controller and broker nodes instead of combined-role nodes.
+         * When enabled, you must also specify the number of controllers using {@link #withNumberOfControllers(int)}.
+         *
+         * @return the current instance of {@code StrimziKafkaClusterBuilder} for method chaining
+         */
+        public StrimziKafkaClusterBuilder withSeparatedRoles() {
```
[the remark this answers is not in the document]
see-quick: It's better to have this explicit method, at least so that the user knows it will use separate roles. I know that `withNumberOfControllers` we can bypass it ... but I am for having this separate method. IMO. Any other opinions @strimzi/maintainers ?

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -157,6 +159,14 @@ protected void doStart() {
             this.imageNameProvider.complete(KafkaVersionService.strimziTestContainerImageName(this.kafkaVersion));
         }
 
+        if (this.nodeRole == KafkaNodeRole.CONTROLLER) {
+            // Controller-only nodes don't need to expose the Kafka client port
+            super.setExposedPorts(Collections.emptyList());
```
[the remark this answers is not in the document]
see-quick: Hmmm, right. We need to expose it here. Forgot to update... I assume we also need to update to the combined mode to expose both ports. Not just the 9092 port.

README.md
```diff
@@ -211,7 +211,22 @@ StrimziKafkaCluster kafkaCluster = new StrimziKafkaCluster.StrimziKafkaClusterBu
 kafkaCluster.start();
 ```
 
-#### xi) Logging Kafka Container/Cluster Output to SLF4J
+#### xi) Run a Kafka cluster with separate controller and broker roles
+
+By default, `StrimziKafkaCluster` uses mixed-role nodes where each node acts as both controller and broker.
```
mimaison, line 216: `mixed` -> `combined`

src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
```diff
@@ -40,28 +40,40 @@ public class StrimziKafkaCluster implements KafkaContainer {
...
     private final String clusterId;
 
     private StrimziKafkaCluster(StrimziKafkaClusterBuilder builder) {
         this.brokersNum = builder.brokersNum;
+        this.controllersNum = builder.controllersNum;
+        this.useSeparatedRoles = builder.useSeparateRoles;
         this.enableSharedNetwork = builder.enableSharedNetwork;
         this.network = this.enableSharedNetwork ? Network.SHARED : Network.newNetwork();
+
+        // internal topics must be <= number of brokers
```
mimaison, line 65: Should it be "replication factor must be <= number of brokers"

src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
```diff
@@ -252,18 +345,29 @@ public Collection<GenericContainer<?>> getNodes() {
...
             .map(KafkaContainer::getBootstrapServers)
             .collect(Collectors.joining(","));
     }
 
+    /**
+     * Get the bootstrap controllers that can be used for controller operations
+     * @return a comma separated list of Kafka controller endpoints
+     */
+    @Override
+    public String getBootstrapControllers() {
```
mimaison, line 365: Should we also have a `getNetworkBootstrapControllers()` method?

src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
```diff
@@ -368,4 +484,45 @@ public void stop() {
...
     }
+
+    /**
+     * Returns the controller nodes.
+     * For combined-role clusters, this returns all nodes.
+     * For separated-role clusters, this returns only the controller-only nodes.
+     *
+     * @return Collection of controller nodes
+     */
+    public Collection<KafkaContainer> getControllerNodes() {
```
mimaison, line 495: Should this be called `getControllers()` for consistency with `getBrokers()`?

src/main/java/io/strimzi/test/container/KafkaContainer.java
```diff
@@ -21,4 +21,15 @@ public interface KafkaContainer extends Startable {
...
+
+    /**
+     * Get the Kafka controller bootstrap servers for admin operations that require controller access.
+     * This is particularly useful in KRaft mode for operations like cluster metadata queries.
+     *
+     * @return controller bootstrap servers
+     * @throws UnsupportedOperationException if the implementation doesn't support controller endpoints
+     */
+    default String getBootstrapControllers() {
+        throw new UnsupportedOperationException("Controller endpoints are not supported by this implementation");
```
mimaison, line 33: Why do we need a default implementation?

src/main/java/io/strimzi/test/container/KafkaNodeRole.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Checks if this node role includes broker functionality.
+     *
+     * @return true if the node can act as a broker, false otherwise
+     */
+    public boolean isBroker() {
+        return this == COMBINED || this == BROKER;
+    }
+}
```
mimaison, line 61: Other files end with a new line. Not sure what's the project's preference but I'd thought I'd mention it

src/test/java/io/strimzi/test/container/KafkaNodeRoleTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class KafkaNodeRoleTest {
+
+    @Test
+    void testMixedRoleProcessRoles() {
```
mimaison, line 15: Still plenty of `mixed` usages in this file

src/test/java/io/strimzi/test/container/KafkaNodeRoleTest.java
```diff
@@ -0,0 +1,58 @@
...
+        assertThat(KafkaNodeRole.BROKER.isController(), is(false));
+    }
+
+    @Test
+    void testMixedRoleIsBroker() {
+        assertThat(KafkaNodeRole.COMBINED.isBroker(), is(true));
+    }
+
+    @Test
+    void testControllerOnlyIsBroker() {
```
mimaison, line 50: We renamed the enum value, should we also update the test names. Same in the few other places in this file.

src/main/java/io/strimzi/test/container/KafkaContainer.java
mimaison, line 1: Should we update the javadoc for `KafkaContainer` as well?

src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
```diff
@@ -85,37 +97,104 @@ private void prepareKafkaCluster(final Map<String, String> additionalKafkaConfig
             defaultKafkaConfigurationForMultiNode.putAll(additionalKafkaConfiguration);
         }
 
-        // multi-node set up
+        if (this.useSeparatedRoles) {
+            prepareSeparateRolesCluster(defaultKafkaConfigurationForMultiNode, kafkaVersion);
```
mimaison, line 101: What about `prepareDedicatedRolesCluster()`? And overall changing "separate(d)" to "dedicated" overall?

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -288,59 +318,77 @@ protected String extractListenerName(String bootstrapServers) {
...
+            bootstrapServers = getBootstrapServers();
+            bsListenerName = extractListenerName(bootstrapServers);
+        }
 
-        // add first PLAINTEXT listener
-        advertisedListeners.append(bootstrapServers);
-        kafkaListeners.append(bsListenerName).append(":").append("//").append("0.0.0.0").append(":").append(KAFKA_PORT).append(",");
-        this.listenerNames.add(bsListenerName);
+        // Only add client listener for broker nodes
+        if (this.nodeRole.isBroker() && bootstrapServers != null && bsListenerName != null) {
```
mimaison, line 336: We don't need `bootstrapServers != null` here as if `this.nodeRole.isBroker()` is true, we've set `bootstrapServers` just above.

mimaison (commented):
Thanks for the updates! I made another pass and left a few more suggestions

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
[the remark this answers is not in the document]
see-quick: removed.

src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
see-quick: Yeah, sounds better.

src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
see-quick: I have implemented it in `getNetworkBootstrapControllers`, you can check.

src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -229,9 +256,12 @@ public StrimziKafkaContainer waitForRunning() {
...
         super.containerIsStarting(containerInfo, reused);
 
-        this.kafkaExposedPort = getMappedPort(KAFKA_PORT);
-
-        LOGGER.info("Mapped port: {}", kafkaExposedPort);
+        if (this.nodeRole.isBroker()) {
+            this.kafkaExposedPort = getMappedPort(KAFKA_PORT);
+            LOGGER.info("Mapped port: {}", kafkaExposedPort);
+        } else {
+            LOGGER.info("Controller-only node - no client port exposed");
```
mimaison, line 263: We don't want to expose `CONTROLLER_PORT`? How would clients interact with the controllers?

see-quick: Ah, we need to allow users to connect outside of the Docker network (I was only thinking clients within the same Docker network). I will update it the same way we do for brokers; you can check the changes. 
---
Also, I added IT, but maybe you could think ofa  better one?

src/test/java/io/strimzi/test/container/StrimziKafkaClusterIT.java
```diff
@@ -232,6 +269,55 @@ private void verifyFunctionalityOfKafkaCluster() throws ExecutionException, Inte
...
+            for (String featureName : expectedFeatures) {
+                if (features.finalizedFeatures().containsKey(featureName)) {
+                    FinalizedVersionRange versionRange = features.finalizedFeatures().get(featureName);
+                    assertThat(versionRange, notNullValue());
+                    LOGGER.info("{}: {}", featureName, versionRange);
+                } else {
+                    LOGGER.warn("Expected feature '{}' not found in finalized features", featureName);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
see-quick: Yeah, I can remove that. Thanks

src/test/java/io/strimzi/test/container/StrimziKafkaClusterIT.java
```diff
@@ -232,6 +269,55 @@ private void verifyFunctionalityOfKafkaCluster() throws ExecutionException, Inte
...
+    }
+
```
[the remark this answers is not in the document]
see-quick: Sure, I can create one.

PR #164, opened 2025-11-05
Fix ToxiProxy integration for multi-broker Kafka cluster

see-quick:
In this PR, I am fixing the behaviour of ToxiProxy. At the current moment, if one were using a Kafka cluster with a proxy (e.g., latencies, bandwidth limitation), such injections would simply not work. So there were a few issues I have addressed in this PR:
1. A race condition occurred when we tried to start the proxy container. 
2. Instead of using a free port, directly use the broker network alias 
3. If using a proxy, then return the correct bootstrap servers
4. and port exposure to proxy ports (also able for external AdminClient/Producer/Consumer) to be able to access those.
5. updating README.md to only mention `StrimziKafkaCluster` instead of  `StrimziKafkaContainer` as such class is now removed from public API.

Also, I have added two simple ITs to cover such functionality and avoid those issues in the future.

### DIFF
--- a/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
+++ b/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
@@ -162,22 +162,6 @@ class StrimziKafkaContainer extends GenericContainer<StrimziKafkaContainer> impl
     @Override
     @DoNotMutate
     protected void doStart() {
-        // Setup proxy container if needed
-        if (this.proxyContainer != null && !this.proxyContainer.isRunning()) {
-            // Synchronize to avoid race conditions when multiple brokers try to start the proxy
-            synchronized (this.proxyContainer) {
-                // Double-check after acquiring lock
-                if (!this.proxyContainer.isRunning()) {
-                    this.proxyContainer.start();
-                }
-            }
-        }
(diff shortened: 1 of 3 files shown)


src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -734,8 +750,10 @@ public String getBootstrapServers() {
         }
         
         if (proxyContainer != null) {
-            // returning the proxy host and port for indirect connection
-            return String.format("PLAINTEXT://%s", getProxy().getListen());
+            this.getProxy();
```
ppatierno: it looks weird to me you are calling a `getXXX` method but actually not using at all what is returned. I can see it's mocking the proxy but why it's not used them? Is `getProxy` the right name for the method?

see-quick: Yes, agree... such a name is a bit misleading. The better way would be to divide the method into 2 separate ones: 
 1. `ensureProxyInitialized()`, which would be a whole mocking proxy stuff...
 2. `getProxy()`, which would just get the proxy as expected

 WDYT?

PR #204, opened 2026-04-13
Refactor input validation by introducing requireNonBlank utility method

### LINKED ISSUE #203: Extract repeated non blank string validation into a utility method
The pattern below is duplicated several times across `StrimziKafkaContainer` and `StrimziKafkaCluster`.
In general, therer is basically this pattern: 
```java
if (value != null && !value.trim().isEmpty()) {
      this.field = value;
  } else {
      throw new IllegalArgumentException("Field cannot be null or empty.");
  }
```
we can simply put this into helper Utils method i.e.,
```java
static String requireNonBlank(String value, String fieldName) {
      if (value == null || value.trim().isEmpty()) {
          throw new IllegalArgumentException(fieldName + " cannot be null or empty.");
      }
      return value.trim();
  }
```
and then basically reduce overall code to just (this is specifically for `saslUsername`):
```java
this.saslUsername = Utils.requireNonBlank(saslUsername, "SASL username");
```

guancioul:
Fixed #203,

Extract repeated non-blank string validation into a utility method

Affected methods                                                                                   

  - StrimziKafkaCluster.StrimziKafkaClusterBuilder#withImage
  - StrimziKafkaCluster.StrimziKafkaClusterBuilder#withLogCollection
  - StrimziKafkaCluster.StrimziKafkaClusterBuilder#withSaslUsername
  - StrimziKafkaCluster.StrimziKafkaClusterBuilder#withSaslPassword
  - StrimziKafkaContainer#withLogCollection
  - StrimziKafkaContainer#withSaslUsername
  - StrimziKafkaContainer#withSaslPassword

### DIFF
--- a/src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
+++ b/src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
@@ -493,11 +493,7 @@ public class StrimziKafkaCluster implements KafkaContainer {
          * @throws IllegalArgumentException if the image name is null or empty
          */
         public StrimziKafkaClusterBuilder withImage(String kafkaImage) {
-            if (kafkaImage != null && !kafkaImage.trim().isEmpty()) {
-                this.kafkaImage = kafkaImage.trim();
-            } else {
-                throw new IllegalArgumentException("Kafka image cannot be null or empty.");
-            }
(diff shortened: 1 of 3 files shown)


see-quick (approved):
LGTM 👍 Thanks @guancioul.

ppatierno (approved):

PR #209, opened 2026-04-14
Fix controller advertised.listeners to use network aliases in dedicated-roles KRaft clusters

see-quick:
Closing #200. Note that everything worked with combined-mode. This is fix change for dedicated roles and also I saw that for consistency getNetworkBootstrapServers and getNetworkBootstrapControllers now returns prefix listener names `CONTROLLER:` or `PLAINTEXT:`.

### DIFF
--- a/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
+++ b/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
@@ -124,8 +124,10 @@ public class StrimziKafkaContainer extends GenericContainer<StrimziKafkaContaine
         CLIENT,
         /** Inter-broker communication listener */
         INTER_BROKER,
-        /** Controller communication listener */
-        CONTROLLER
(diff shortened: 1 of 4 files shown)


im-konge (approved):

see-quick:
Tests don't run as a part of CI, which I realised ... (and I think here I would need more thinking because some of the tests are not passing locally :/ so that's why I am labeling this do-no-merge)

Frawless:
> Tests don't run as a part of CI, which I realised ... (and I think here I would need more thinking because some of the tests are not passing locally :/ so that's why I am labeling this do-no-merge)

Next rerun should run the tests.

see-quick:
As also discussed with other folks we should go with having two listeners (i.e., CONTROLLER for internal docker network, CONTROLLER-EXTERNAL for external access for clients). Two listeners would be only created only in separated/dedicated mode (i.e., in combined mode we will have only CONTROLLER listener).

im-konge (approved):

PR #210, opened 2026-04-17
Lower pitest line coverage threshold from 76 to 75

see-quick:
In the previous PR (i.e., https://github.com/strimzi/test-container/pull/204), we refactored inline validation logic from Strimzi* classes into Utils class but that class is explicitly excluded from the pitest target classes scope.

Which eventually means that 14 LoC of moved out of coverage => causing coverage to drop (i.e.,  from 76 to 75). Even we didn't loss any test coverage. Mutation score is still 100% anyway but with these changes I am thinking if we should include f.e., Utils method or others also into account so we would avoid these problems in the future.

WDYT?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -418,7 +418,7 @@
                         <avoidCallsTo>org.apache.commons.logging</avoidCallsTo>
                     </avoidCallsTo>
                     <mutationThreshold>100</mutationThreshold>
-                    <coverageThreshold>76</coverageThreshold>
+                    <coverageThreshold>75</coverageThreshold>
                     <verbose>true</verbose>
                 </configuration>
             </plugin>


scholzj (approved):
Don't get me wrong ... I'm completely fine with you merging this and I have approved it.

But what is the value of a tool like this that does not capture all files and where we lower the requirement because we moved something somewhere else? (already the original value 76 seems rather suspitiously _random_)

see-quick:
> Don't get me wrong ... I'm completely fine with you merging this and I have approved it.
> 
> But what is the value of a tool like this that does not capture all files and where we lower the requirement because we moved something somewhere else? (already the original value 76 seems rather suspitiously _random_)

Yeah thats why I raise this https://github.com/strimzi/test-container/issues/211. I would like to extend it to all classes IMHO, as I didn't like expected that this could be a problem (having just specific core classes in the coverage) in the future (my bad I guess).

im-konge (approved):
I'm fine with changing it, but I think I had the same comment in the previous PR - should it be really changed? Would it make sense to have it more "voluntary" and for example run some profile that will just show you how much coverage you have? IDK if you have some checks and pipelines connected to this, but maybe it's something that would fix this thing with changing the value. Or can you just remove the coverage threshold completely? So it will just show you the percentage?

see-quick:
The main purpose of this mandatory check is that if you change code and do not write any tests then you will be warned in the build (f.e., you have 100 LoC and all 100 are covered you have 100/100 => 100% but if you add method which has 15 LoC and not add any test you end-up with 100/115  => ~86% => violation) => so you would be pushed to write tests.

The problem with current situation is that if you move some code to classes which are not specified in the target classes then it basically removes that coverage (which is not good) and I think it should be changed that we would have all classes covered and that way we would also avoid this issue which we currently have.

scholzj:
Why not use Codecov as in Operators?

see-quick:
Because I don't know codecov as much as I do pitest. And AFAIK codecov is just doing only line/branch coverage which is not a replacement for pitest and it basically tells you which lines were executed nothing more (during tests). With mutation coverage you have also another dimension of mutants (e.g., flipping operators, functions return types) and for each generated test case (i.e., mutation) we check each if any fails. If no test catches the spawned mutation (i.e., wrong flipped operator) it means the test suite has a gap that line coverage would never reveal :). And maybe codecov has some plugin which also has mutation coverage but I think if this works why should I change to codecov? Only need would be to just have all classes to scan instead specific ones as we do currently and we will avoid this kind of problems (i.e., moving some part of code into filtered classes). Only one thing where codecov is beneficial is dashboarding ... but I think I can live without it. 

And if having this strict coverage threadhold is problem ... then we can have it similarly as we do in operators (i.e., 50%). Although I am more to have still this strict threshold so if you change code then you will be warned with that and you will need to add tests to cover such functionality.

PR #221, opened 2026-05-04
[Chore] Remove pitest targetClasses restriction and lower thresholds

### LINKED ISSUE #217: Remove pitest targetClasses restriction
Remote pitest targetClasses restriction in pom.xml and then lower coverageThreshold and mutationThreshold as needed.  Keep targetTests as `*Test` (no ITs included only mockUTs and UTs).

q404365631:
## Changes

- Remove `<targetClasses>` restriction to allow mutation testing on all classes (WaitException, Utils, KafkaVersionService, etc.)
- Lower `mutationThreshold` from 100 → 60
- Lower `coverageThreshold` from 75 → 50
- Keep `targetTests` as `*Test` (no integration tests)

Closes #217

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -401,12 +401,6 @@
                     </dependency>
                 </dependencies>
                 <configuration>
-                    <targetClasses>
-                        <!-- we not want to include also WaitException, Utils, KafkaVersionService because we mainly
-                            focus mutation core classes (i.e., Strimzi*, KafkaContainer) -->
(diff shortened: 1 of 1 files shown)


see-quick:
Please resolve the DCO (i.e., you need to sign the commits using `-s`) and also I would like to see before and after screenshot of those classes so you can add that inside the PR description. You can find those parts inside `./target/pit-reports/index.html`.

Issue #214, opened 2026-04-30
Add Kafka 4.3.0 support


see-quick:
We should add Kafka 4.3.0 support once Kafka released it. You can follow this PR https://github.com/strimzi/test-container/issues/198.

You can watch here https://dlcdn.apache.org/kafka/ to confirm that such version of Kafka is present.

bboyleonp666:
@see-quick i wonder if we should bump the kafka version in _pom.xml_ as well?

PR #225, opened 2026-05-23
Add Kafka 4.3.0

bboyleonp666:
Address https://github.com/strimzi/test-container/issues/214.

Add Kafka 4.3.0 image support.

### DIFF
--- a/src/test/java/io/strimzi/test/container/StrimziKafkaClusterIT.java
+++ b/src/test/java/io/strimzi/test/container/StrimziKafkaClusterIT.java
@@ -1136,48 +1136,12 @@ public class StrimziKafkaClusterIT extends AbstractIT {
             Transferable.of(systemUnderTest.getClientTrustStoreBytes()),
             "/tmp/client.truststore.p12");
 
-        // The internal listener only trusts the cluster CA. The client cert is signed by the
-        // clients CA, so the handshake fails. Under TLS 1.3, the client checks the server's
-        // CertificateRequest CA list, finds no match, and sends an empty Certificate message.
-        // The server responds with a certificate_required alert (RFC 8446 §6.2).
-        assertClientCertRejectedByInternalListener(broker, systemUnderTest.getClientStorePassword(),
-            "TLSv1.3", "certificate_required");
-    }
+        String password = systemUnderTest.getClientStorePassword();
 
-    @Test
-    void testClientCertRejectedByInternalListenerWithTls12() throws Exception {
-        systemUnderTest = new StrimziKafkaCluster.StrimziKafkaClusterBuilder()
-            .withNumberOfBrokers(NUMBER_OF_REPLICAS)
-            .withTls()
-            .build();
-
-        systemUnderTest.start();
-
-        StrimziKafkaContainer broker = systemUnderTest.getNodes().iterator().next();
-
-        // Copy client keystore and truststore into the container
-        broker.copyFileToContainer(
-            Transferable.of(systemUnderTest.getClientKeyStoreBytes()),
-            "/tmp/client.keystore.p12");
-        broker.copyFileToContainer(
-            Transferable.of(systemUnderTest.getClientTrustStoreBytes()),
-            "/tmp/client.truststore.p12");
-
-        // The internal listener only trusts the cluster CA. The client cert is signed by the
-        // clients CA, so the handshake fails. Under TLS 1.2, the client checks the server's
-        // CertificateRequest CA list, finds no match, and sends an empty Certificate message.
-        // The server responds with a handshake_failure alert (RFC 5246 §7.2.2).
-        assertClientCertRejectedByInternalListener(broker, systemUnderTest.getClientStorePassword(),
-            "TLSv1.2", "handshake_failure");
-    }
-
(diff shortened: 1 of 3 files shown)


src/test/java/io/strimzi/test/container/StrimziKafkaClusterIT.java
```diff
@@ -1161,16 +1162,75 @@ void testClientCertRejectedByInternalListener() throws Exception {
...
+            "/tmp/client.keystore.p12");
+        broker.copyFileToContainer(
+            Transferable.of(systemUnderTest.getClientTrustStoreBytes()),
+            "/tmp/client.truststore.p12");
+
+        String password = systemUnderTest.getClientStorePassword();
+
+        // Force TLS 1.2: the client checks the server's CertificateRequest CA list, finds no
+        // match, and sends an empty Certificate message.
+        String commandConfig = String.join("\\n",
```
chia7712, line 1143: Should we introduce a helper method to eliminate the duplicate code?

bboyleonp666: awesome idea. update in [d1c14ab](https://github.com/strimzi/test-container/pull/225/commits/d1c14ab395fafcb878122197a2a4f9caa555cb0c)

src/test/java/io/strimzi/test/container/StrimziKafkaClusterIT.java
```diff
@@ -1136,12 +1136,48 @@ void testClientCertRejectedByInternalListener() throws Exception {
...
+        // The internal listener only trusts the cluster CA. The client cert is signed by the
+        // clients CA, so the handshake fails. Under TLS 1.3, the client checks the server's
+        // CertificateRequest CA list, finds no match, and sends an empty Certificate message.
+        // The server responds with a certificate_required alert (RFC 8446 §6.2).
+        assertClientCertRejectedByInternalListener(broker, systemUnderTest.getClientStorePassword(),
+            "TLSv1.3", "certificate_required");
+    }
+
+    @Test
+    void testClientCertRejectedByInternalListenerWithTls12() throws Exception {
```
see-quick: I think we do not need to check every version of TLS, especially since this may change over time again? So I would suggest focusing on the latest one (i.e., `TLSv1.3`). WDYT @strimzi/maintainers ?

see-quick (commented):
Interesting, thanks for checking.

src/test/java/io/strimzi/test/container/StrimziKafkaClusterIT.java
bboyleonp666: i am good with that. as long as the TLS wording are valid for both TLS v1.2 and v1.3 (verified on my local), there's no need to introduce a new test which increases the CI burden. besides, it's just a wording check, not affecting the overall behavior.
let's see if there are some other voices.

scholzj: > I think we do not need to check every version of TLS, especially since this may change over time again? So I would suggest focusing on the latest one (i.e., `TLSv1.3`). WDYT @strimzi/maintainers ?

TBH, I'm not sure I follow the problem or the solution. And I absolutely don't understand how it relates to Kafka 4.3.0.

If we want to check the exact error message and they might differ, we could use some qualifiers in the asserts etc. Maybe we also do not need to check the exact message given it is not our message in the first place.

bboyleonp666: > I absolutely don't understand how it relates to Kafka 4.3.0.

@scholzj You are right. That message has nothing to do with Kafka 4.3.0, not even strimzi project.

@see-quick As of now, both TLS versions share the same cert reject behavior, just differs in the messages. I think removing the message checking in the test and focusing on the TLS cert reject behavior is the goal. That way, we don't need the test for TLS v1.2 as well. WDYT?

see-quick: Good point, lets keep only the TLS 1.3 test case, drop the alert message assert and just assert on `SslAuthenticationException.`.

see-quick: but I would rather create separate PR for this, as Jakub pointed out above so we have better tracking and we are not hiding this problem within Adding Kafka 4.3.0 version (which is totally un-related). Thanks.

bboyleonp666: done. do we normally open an issue to do the follow-up? that could be a good-first lol.

see-quick: This is simple one, so just open a separate PR with the fix :).

see-quick: Created to not block other PRs https://github.com/strimzi/test-container/pull/228 and avoid build errors.

bboyleonp666: thanks for that @see-quick . i will rebase my change once merged.

see-quick (approved):
Thanks @bboyleonp666 👍 .

scholzj (approved):

PR #226, opened 2026-05-26
Add method to retrieve latest patch versions for Kafka minor releases

### LINKED ISSUE #215: Add `getLatestPatchVersions()` method to StrimziKafkaCluster
Currently `getSupportedKafkaVersions()` returns all supported Kafka versions. Multiple projects (metrics-reporter, bridge) need to run integration tests only against the latest patch release of each minor version (e.g. 4.0.1, 4.1.1 instead of all versions).

Adding a `getLatestPatchVersions()` method to StrimziKafkaCluster would allow these projects to share the filtering logic via `@MethodSource`, rather than each project implementing its own filtering utility.

Related: https://github.com/strimzi/metrics-reporter/issues/107

guancioul:
Closes #215

### Summary
  
- Add `getLatestPatchVersions()` to `StrimziKafkaCluster` as a public API, which returns the latest patch release for each minor version (e.g. [3.7.1, 4.0.1, 4.1.0])
- Extract the core filtering logic into a package-private static method in `KafkaVersionService` to allow unit testing with controlled input
- Add unit test with a fixed set of versions to verify that only the latest patch per minor is returned

### Test plan

- `KafkaVersionServiceTest#testGetLatestPatchVersionsHasOneVersionPerMinor` — verifies the filtering logic with a controlled input list, independent of the real `kafka_versions.json`

### DIFF
--- a/src/main/java/io/strimzi/test/container/KafkaVersionService.java
+++ b/src/main/java/io/strimzi/test/container/KafkaVersionService.java
@@ -201,6 +201,32 @@ class KafkaVersionService {
     }
 
     /**
+     * Get list of the latest patch release for each minor version.
+     * For example, if supported versions are [3.7.0, 3.7.1, 4.0.0, 4.0.1, 4.1.0],
+     * this returns [3.7.1, 4.0.1, 4.1.0].
+     *
+     * @return list of version strings representing the latest patch per minor, sorted from oldest to newest
+     */
+    public List<String> getLatestPatchVersions() {
+        return getLatestPatchVersions(this.logicalKafkaVersionEntities);
+    }
+
+    static List<String> getLatestPatchVersions(List<KafkaVersion> versions) {
+        return versions.stream()
+            .collect(Collectors.toMap(
+                kv -> {
+                    String[] parts = kv.getVersion().split("\\.");
+                    return parts[0] + "." + parts[1];
+                },
+                KafkaVersion::getVersion,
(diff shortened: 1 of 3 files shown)


src/main/java/io/strimzi/test/container/KafkaVersionService.java
```diff
@@ -200,6 +200,32 @@ public List<String> getSupportedKafkaVersions() {
...
+    /**
+     * Get list of the latest patch release for each minor version.
+     * For example, if supported versions are [3.7.0, 3.7.1, 4.0.0, 4.0.1, 4.1.0],
+     * this returns [3.7.1, 4.0.1, 4.1.0].
+     *
+     * @return list of version strings representing the latest patch per minor, sorted from oldest to newest
+     */
+    public List<String> getLatestPatchVersions() {
+        return getLatestPatchVersions(this.logicalKafkaVersionEntities);
+    }
```
see-quick, line 212: I would also add test case for this method where you call `KafkaVersionService.getInstance().getLatestPatchVersions()` and then you verify that versions are not null and empty (without defining it explicitly).

see-quick (commented):
Looks good, thanks. Just one small nit.

src/main/java/io/strimzi/test/container/KafkaVersionService.java
guancioul: Sure, I think that makes sense to check this!

see-quick (approved):
Thanks @guancioul 👍

im-konge (approved):
LGTM, thanks :)

PR #227, opened 2026-05-27
Refactor ListenerConfig and ListenerRole into Listener class with constants

### LINKED ISSUE #212: Extract ListenerConfig/ListenerRole into separate classes and define listener name constants
Currently the main class (i.e., StrimziKafkaContainer) is growing and we should separate it a bit. One of such example is to extract ListenerConfig/ListenerRole classes. 

If we extract this and define it as separate class it would be also great to define our all listeners names. I imagine it like:
```java
class Listener {

      enum Role {
          CLIENT,              // external communcation 
          INTER_BROKER,        // internal broker
          CONTROLLER.          // ...
      }
      
      // then here we define a constants 
      static final String CONTROLLER = "CONTROLLER"; 
      static final String PLAINTEXT = "PLAINTEXT";
      // ... (etc.)
      
      // field for name 
      // field for role 
      
     // ... and constructor and related methods
```

that way we reduce number of magic strings across many methods (e..g, buildListenersConfig, setKRaftProperties ...). Keep it more readable across whole class.

> [!NOTE] 
> We should keep them internal (i.e., package-private) and maybe in the future we could consider to expose them via public API and if we go with custom listener API.

guancioul:
Close #212,

### Summary

- Extract the `ListenerConfig` inner class and `ListenerRole` enum from `StrimziKafkaContainer` into a new top-level `Listener` class (`Listener.java`), reducing the coupling between listener configuration and the container class
- Define listener name constants (`PLAINTEXT`, `CONTROLLER`, `CONTROLLER_EXTERNAL`, `INTER_BROKER_PREFIX`) in `Listener` to eliminate magic strings scattered across `StrimziKafkaContainer`
- Update all call sites in `StrimziKafkaContainer` and `StrimziKafkaContainerTest` to use the new `Listener` class and its constants


see-quick (commented):
Thanks for this change @guancioul, please also update all test suites which uses exact strings of the listeners and change it to Listener.PLAINTEXT ...  (i.e.,`StrimziKafkaContainerMockTest`, `StrimziKafkaClusterTest`, `StrimziKafkaContainerIT`).

see-quick (approved):
LGTM, thanks 👍

Frawless (approved):

PR #228, opened 2026-05-27
Remove specific assertion from the TLS test case

see-quick:
We only assert on SslAuthenticationException (not the specific alert string) because [JDK-8311644](https://bugs.openjdk.org/browse/JDK-8311644) changed the TLS alert from `bad_certificate` to the RFC-correct value, and the exact wording differs across JDK versions and TLS protocol versions.


bboyleonp666 (approved):
LGTM

scholzj (approved):

guancioul (approved):
LGTM, Thanks!

PR #232, opened 2026-06-04
Add release_json target to improve release proces and avoid setting it manually but via release job

see-quick:
This PR adds additional step to patch our `kafka_versions.json` so we don't need to do it manually but via release job.


scholzj (commented):
Doesn't the script need to be immutable? E.g. for patch releases etc.? (+ I guess is it immutable or not? It is not clear to me without trying it)

scholzj (approved):

ppatierno (approved):

PR #233, opened 2026-06-05
Skip ITs on release branches

see-quick:
### Type of change

- Task

### Description

This PR adds a condition where to run ITs. Currently we run it also on release branches but that would fail because such images are not present (GAs ones) as we are doing release candidates.

We can't run ITs in the `release-*` branches as we build images in separate repository.

### Checklist

- [x] Update documentation
- [ ] Update CHANGELOG.md (if present)
- [ ] Reference relevant issue(s) and close them after merging
- [ ] Write tests
- [ ] Make sure all tests pass
- [ ] Try your changes inside a Kubernetes cluster, not just from unit tests
- [ ] AI assistance was used to create this PR

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,12 +36,18 @@ jobs:
         with:
           javaVersion: ${{ matrix.java.version }}
 
+      # On release branches, skip ITs and run only unit tests and mutation testing.
+      # ITs cannot run here because release candidates use -rc1 suffix tags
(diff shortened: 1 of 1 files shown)


scholzj (commented):
Doens't that suggest some issue in the build that should be fixed rather then skip the ITs?

Frawless (approved):

scholzj (approved):

PR #234, opened 2026-06-08
Include mutation testing in the test scope

see-quick:
### Type of change

- Bugfix
- Enhancement / new feature

### Description

This PR adds mutation testing in the test scope as it seems that when migrating to GHA, we forgot to add that. By default it run only Azure pipelines. This should solve this as it would run after UTs so the flow would be:

UTs -> mutation tests -> ITs.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md (if present)
- [ ] Reference relevant issue(s) and close them after merging
- [ ] Write tests
- [x] Make sure all tests pass
- [ ] Try your changes inside a Kubernetes cluster, not just from unit tests
- [ ] AI assistance was used to create this PR (see the [Strimzi AI policy](https://github.com/strimzi/governance/blob/main/AI_POLICY.md))

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -400,6 +400,15 @@
                         <version>${pit-junit-plugin.version}</version>
                     </dependency>
                 </dependencies>
+                <executions>
+                    <execution>
+                        <id>mutation-testing</id>
+                        <phase>test</phase>
+                        <goals>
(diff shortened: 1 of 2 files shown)


Frawless (approved):
Assuming it won't break anything

